PR #214, opened 2014-04-25
Remove trailing whitespace from docs.

bk2204:
Trailing whitespace is widely considered a bad idea, so this commit removes it from all the AsciiDoc files in the docs directory.

This change only affects the docs directory to keep the amount of changes something I could reasonably review. The only difference when processing the docs before and after this change is the last updated timestamp, so this change has no functional impact.

The command I used to perform this was:

``` sh
cd docs; for i in $(git grep -lP '\s+$'); do [[ -z ${i%%*.pdf} ]] && continue;  ruby -pi -e '$_.gsub!(/\s+$/, "\n")' $i; done
```

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+[*]
+charset = utf-8
+end_of_line = lf
+
+[*.adoc]
+indent_size = 2
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
--- a/README.adoc
+++ b/README.adoc
@@ -96,7 +96,7 @@ First, install http://rvm.io[RVM]. Trust me on this.
 
 Next, setup an RVM gemset for working with the Asciidoctor site:
 
- $ rvm use 2.3.3@asciidoctor.org --create 
+ $ rvm use 2.3.3@asciidoctor.org --create
 
 At this point, you could install Awestruct directly, but since the site build has some additional dependencies, it's best to let bundler handle the installation. Bundler will also ensure that you are using the correct versions of each gem when you run Awestruct.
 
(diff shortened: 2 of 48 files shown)


olleolleolle:
Hi! Can we resolve this issue? 

- Closing ot without change?
- Teaching users about settings in editors by adding meta-docs (i.e. document how to author documentation)?
- Adding a pre-commit Git script one must run?
- Adding a Travis build rule that fails edits to `.adoc` files when they have whitespace at end-of-line?

mojavelinux:
What needs to happen here first is that this change needs to be reapplied. If someone wants to run with it, feel free to submit a new PR.

I completely agree that moving forward we should have:

* .editorconfig file in the root of the project with the shared settings
* a validation on the CI job that runs* to ensure no trailing whitespace is added

\* CI runs the `travis` rake task. We could add a validate task to the Rake build that checks for trailing whitespace and throws an exception if it is detected. The failure routine will catch it and flag the PR.

bk2204:
I've adjusted the commit to remove all the existing trailing whitespace in *.adoc files.  Additionally, I've made the travis target raise an exception on trailing whitespace and added a .editorconfig covering *.adoc.

I wasn't sure if we wanted to cover some settings more globally.  For example, we probably want to require UTF-8 for all files, as well as using LF for line endings.  Let me know what changes you want, and I'll resubmit.

mojavelinux:
Thank you @bk2204!

> we probably want to require UTF-8 for all files, as well as using LF for line endings. 

:+1: to both of these validations.

mojavelinux:
Let's go ahead and merge this and handle the other enhancements in a follow-up PR.

mojavelinux:
Thanks @bk2204!

PR #254, opened 2014-08-08
Update install-upgrade-asciidoctor.adoc

alacerda:
Added the Alpine Linux installation (apk) instruction


mojavelinux:
Where I think this change is going to fit is in the planned Ruby environment setup guide (#399).

Once Ruby is installed, the instructions to install Asciidoctor should be the same on all systems (assuming the user is taking the `gem install` approach).

mojavelinux:
Actually, it seems like there is now a package on Alpine Linux for Asciidoctor, so this will fit right in next to Debian and Fedora.

```
$ sudo apk add asciidoctor
```

mojavelinux:
Merged a revised version of this PR as 349aca9b48109e679dc2558b1a21f7ae10b57d53.

mojavelinux:
Thanks!

PR #331, opened 2014-12-03
Document that we support Ruby 2.1.

bk2204:
There were two places we didn't specify that we supported 2.1.  I chose 2.1._x_ because 2.1 versions are coming out rapidly enough that I didn't want to hard-code 2.1.5 and then have to fix it later.  Let me know if you'd prefer just "2.1" or something else entirely.

### DIFF
--- a/docs/_includes/install-upgrade-asciidoctor.adoc
+++ b/docs/_includes/install-upgrade-asciidoctor.adoc
@@ -11,7 +11,8 @@ Asciidoctor requires one of the following implementations of Ruby:
 
 * Ruby 1.8.7
 * Ruby 1.9.3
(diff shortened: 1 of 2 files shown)


mojavelinux:
I finally got around to merging a revised version of this PR. I documented it as follows:

* Ruby 2 (2.0.0 or better)

We're chasing Ruby versions a bit, but we test forward on all versions. So I think simply saying Ruby 2 is enough so long as we explicitly state that it's everything forward from that version.

mojavelinux:
I also dropped the minor version on JRuby 1.7 since that's just noise and likely to be out of date very quickly.

PR #336, opened 2015-01-02
Use actual attribute name in example

kytrinyx:


mojavelinux:
Thanks @kytrinyx! That definitely gets the point across better.

PR #337, opened 2015-01-04
Add documentation for citetitle substitutions.

kytrinyx:
This mostly a starting point for discussion—I'm not quite sure if I've put things in the right place, or worded these things clearly.

See https://github.com/asciidoctor/asciidoctor/issues/1254

### DIFF
--- a/docs/asciidoc-writers-guide.adoc
+++ b/docs/asciidoc-writers-guide.adoc
@@ -820,6 +820,11 @@ h|Verbatim |Yes                |Yes      |No     |No         |No           |No  
 h|None     |No                 |No       |No     |No         |No           |No     |No
 |===
 
+
+In order to apply normal substitutions to an attribute value, surround it with single quotes.
+There are two exceptions to this behavior: At the moment normal substitutions are not applied to
+the `options` and `title` attribute values.
+
 You can control how blocks are displayed using block metadata.
 
 === Block metadata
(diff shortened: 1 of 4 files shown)


docs/asciidoc-writers-guide.adoc
```diff
@@ -820,6 +820,9 @@ h|Verbatim |Yes                |Yes      |No     |No         |No           |No
 h|None     |No                 |No       |No     |No         |No           |No     |No
 |===
 
+
+In order to apply normal substitutions to the `citetitle` of a quote block you must surround the attribute value with single quotes.
```
mojavelinux: I went back to the source and it turns out that any block attribute value (explicit: `key='value'` or positional: `'value'`) is subject to normal substitutions except for the `options` and `title` attributes. The `options` attribute value never gets substituted. The `title` attribute value always receives title substitutions.

Here's the relevant section of the code.

* https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/attribute_list.rb#L160-L182

There are some edge cases, but we can at least start by documenting as: single quotes mean "apply normal subs".

Aside from this change request, :+1: to the update. Thanks!

kytrinyx:
I've updated the patch to make the suggestions more general.

mojavelinux:
Solid! Thanks @kytrinyx!

PR #349, opened 2015-01-25
Section Titles (Headers -> Headings)

blablatros:
(very) minor change

Section titles are wrongly called headers instead of headings. This change will remove a potential ambiguity between "document header" and "headers", and make it consistent with the rest of the text (ex. in the Markdown Compatibility section where the term headings is being used).


mojavelinux:
I really like this clarification. It's also consistent with the HTML5 specification. See http://www.w3.org/TR/html5/sections.html#headings-and-sections

mojavelinux:
Thank you!

PR #353, opened 2015-02-06
Update the download link of the Firefox addon

Mogztter:
Version 0.3.0 is available with Asciidoctor 1.5.2.js and Opal 0.6.3


mojavelinux:
Nice!

PR #355, opened 2015-02-15
Table attribute 'grid' works just fine

grdryn:
All values for grid appear functional: all, cols, rows, none.

### DIFF
--- a/docs/_includes/sum-table.adoc
+++ b/docs/_includes/sum-table.adoc
@@ -47,7 +47,7 @@ d|`{vbar}`, `:`, `,`, `!`, or a user defined attribute
 .4+|draws boundary lines between rows and columns
 |all
 |default, draws boundary lines around each cell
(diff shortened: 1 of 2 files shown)


grdryn:
I realise that there might be situations that I don't know about, that make it 'not fully implemented' or something. Just PR'ing in case that's not the case :)

mojavelinux:
@grdryn thanks for pointing this out! Indeed, one of the enhancements that came with the 1.5.0 release was proper support for the grid and frame in the default stylesheet. Therefore, we can definitely say that this is fully functional now. Thanks!

Btw, I'm pretty sure we do support all the align attributes. Only the float is not implemented to my knowledge...if you want to keep updating.

grdryn:
@mojavelinux some small testing seems to show that 'align' isn't supported, but 'float' is. I'll keep playing with it, and I'll submit another PR to update what works and what doesn't.

PR #356, opened 2015-02-15
Attr warn

plaindocs:
Fix for asciidoctor/asciidoctor#1277


mojavelinux:
Thanks! I love these types of fixes because it tells people about features they don't even know they had :)

PR #357, opened 2015-02-19
Fix Guard and Bundler usage

j-manu:
* `rb-inotify` is Linux only. Guard now uses `listen` gem which will works cross platform
* It is recommended to use Bundler through `bundle` binary instead of `bundler`
* When using bundler, the Guardfile needs `Bundler.require(:default)`


mojavelinux:
Most excellent! Thanks for making these updates, @j-manu!

PR #360, opened 2015-03-01
Table attribute float works now

grdryn:


grdryn:
This is in response to conversation in #355. I could not see any effect of align, halign, or valign; either visually, or in the generated HTML.

mojavelinux:
That's correct. I think I only did float because it was the only CSS-like property. The rest are relics from HTML4. Thanks for looking into this and making the update, Gerard!

PR #365, opened 2015-03-20
Fixed overlooked change when copypasting

hijarian:
Hello! I see that someone copypasted the line of text without changing the topic. This is a fix for it.


mojavelinux:
Thanks @hijarian!

PR #368, opened 2015-03-26
Change Links to relative files

jmini:
As discussed on the mailing list: [Link to an other document ](http://discuss.asciidoctor.org/Link-to-an-other-document-td2857.html) the section [Links to relative files](http://asciidoctor.org/docs/asciidoc-writers-guide/#links-to-relative-files) of the "AsciiDoc Writer’s Guide" needs to be updated.

### DIFF
--- a/docs/_includes/ex-url.adoc
+++ b/docs/_includes/ex-url.adoc
@@ -49,11 +49,11 @@ Chat with other Asciidoctor users in the irc://irc.freenode.org/#asciidoctor[Asc
 // end::css[]
 
 // tag::link[]
-link:editing-asciidoc-with-live-preview[Editing with Live Preview]
+link:protocol.json[Open the JSON file]
 // end::link[]
 
 // tag::hash[]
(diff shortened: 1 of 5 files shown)


mojavelinux:
I'd like to propose a modification to this change. It's important that we still cover the link macro because it is useful in other contexts (linking to a json file, for instance).

What we should make clear in the existing section about relative links is that *if* you are linking to another AsciiDoc document, the the preferred way to specify that link is to use the cross reference syntax. Perhaps we can call addition to it using a TIP admonition block.

Could you update the pull request accordingly? Thanks!

jmini:
I have made an other proposal in 8d8db07b2a5bcad22339b56bc7ad8b8a1be75c7f. Please tell me what you think. 

Other question: should I rebase on head or not?

mojavelinux:
This looks fantastic! I'm having @graphitefriction look it over first, then we'll merge once she gives the thumbs up.

graphitefriction:
Thank you @jmini ! The clarifications to the text and new examples look great.

PR #376, opened 2015-04-22
Update sum-header.adoc

seanosterberg:
Updated to add keywords to the table


mojavelinux:
Thanks @seanosterberg!

PR #377, opened 2015-05-01
Precision for ordered list

jmini:
As discussed on the [mailing list](http://discuss.asciidoctor.org/Number-lists-give-warnings-td2526.html), ordered list using number needs to be sequential.

This pull request adds more information in the ordered list section.


mojavelinux:
Fantastic! Thanks @jmini!

PR #378, opened 2015-05-01
Fix include with multiple ranges example

jmini:
Related discussion on the [mailing list](http://discuss.asciidoctor.org/Include-directive-by-line-range-with-multiple-ranges-not-working-tp3056.html)

### DIFF
--- a/docs/_includes/ex-include.adoc
+++ b/docs/_includes/ex-include.adoc
@@ -22,8 +22,12 @@ This is documentation for project X.
 \include::filename.txt[lines=5..10]
 // end::line[]
 
+// tag::m-line-comma[]
+\include::filename.txt[lines="1..10,15..20"]
+// end::m-line-comma[]
+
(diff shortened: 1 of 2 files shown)


mojavelinux:
Stellar!

How about adding the following note (either as a paragraph or admonition note...though I think paragraph is probably fine).

> When including multiple line ranges or multiple tags, the individual values can be separated either by a comma or a semi-colon. If commas are used, then the entire attribute value must be enclosed in quotes. Using the semi-colon as the data separator alleviates this requirement.

(You can just update the pull request if you would like).

jmini:
New proposal with your suggested paragraph and additional example (it took me some time to understand what you meant with this quote stuff).

Commit is amended and rebased.

mojavelinux:
Great work! These changes are certainly going to clear up a lot of confusion. Thanks!

PR #392, opened 2015-06-01
Added in gem alpha build

jaredmorgs:
Added in some details about the Alpha gem just released, and broke procedures down into gem and script. :D

### DIFF
--- a/docs/convert-asciidoc-to-pdf.adoc
+++ b/docs/convert-asciidoc-to-pdf.adoc
@@ -89,7 +89,7 @@ If you're using {uri-rvm}[RVM], we recommend creating a new gemset to work with 
 RVM keeps the dependencies required by various projects isolated.
 
 == Install Using +gem+
-You can download an Alpha version of {project-name} using +gem+. The Alpha works for simple AsciiDoc documents, but can not currently traverse `include` statements, nor process TOC macros. See the {uri-worklog}[WORKLOG] for the full list of features planned for inclusion in future releases.
+You can download an Alpha version of {project-name} using +gem+.
 
 .How To Install asciidoctor-pdf Using +gem+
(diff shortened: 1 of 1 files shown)


jaredmorgs:
Just fixed up the statement about the gem build not being able to traverse includes. We fixed that up in my doc so it works correctly.

mojavelinux:
@jaredmorgs What I'm about to tell you is going to seem a bit bizarre, so I apologize in advance. The pages on asciidoctor.org about the individual components such as Asciidoctor PDF are actually a copy of the README.adoc from the project. I usually try to make the changes to the README first, then sync those changes into asciidoctor.org.

The section you added had already been added to the upstream README. However, you may still want to apply revisions. Now that the files are synced, could you look at the README and see if you want to make any updates to the Getting started section?

https://github.com/asciidoctor/asciidoctor-pdf/blob/master/README.adoc

On a side note, I'm well aware that we need to automate this process so that the website sources the content about the individual components like Asciidoctor PDF from the project README (or at least grabs part of it). The current approach is not sustainable...but we haven't had time to make it right.

jaredmorgs:
This is all looking great to me, Dan. Like the way you worded the Gem stuff. 

Would using something like JAQ work with the Awestruct framework?

mojavelinux:
:+1: 

> Would using something like JAQ work with the Awestruct framework?

Absolutely. In fact, I'd even say more so because the whole idea of Awestruct from the beginning was to provide more extensibility thank Jekyll. That's a large reason why Red Hat selected it (aside from it being developed by employees).

mojavelinux:
I recommend talking to Aslak Knutsen or Pete Muir about the things that they've been able to do with Awestruct because it's pretty wild how much automation they've built into it.

PR #393, opened 2015-06-05
Update Twitter URL

adam820:
Update the Twitter URL, as the existing one does not direct to the account page.


mojavelinux:
Thanks @adam820!

PR #411, opened 2015-08-13
Fix typo on homepage

mattcone:
Let's change "lead" to "led."


mojavelinux:
Doh! Thanks for catching this!

PR #415, opened 2015-08-31
Added documentation of iconsfont-name and iconsfont-remote

abelsromero:
In relation to issue https://github.com/asciidoctor/asciidoctor-maven-plugin/issues/167, documentation about how to set FontAwesome offline has been added.

### DIFF
--- a/docs/_includes/icons.adoc
+++ b/docs/_includes/icons.adoc
@@ -53,13 +53,21 @@ This is how the admonition looks rendered.
 NOTE: Asciidoctor supports font-based admonition icons, powered by Font Awesome!
 ====
 
-Asciidoctor adds a reference to the Font Awesome stylesheet and font files served from a CDN to the document header:
+Asciidoctor adds a reference to the Font Awesome stylesheet and font files served from a https://cdnjs.com/[CDN] to the document header:
 
 [source,xml]
 ----
 <link rel="stylesheet"
   href="http://cdnjs.cloudflare.com/ajax/libs/font-awesome/3.1.0/css/font-awesome.min.css">
 ----
+
+In case the stylesheet and fonts must be available offline, it is possible to set as reference a folder path relative to the document.
+
+To do so, set the attributes `iconfont-remote` to `false`, and `iconfont-name` with the path to Font Awesome stylessheet.
+
+ $ asciidoctor -a iconfont-remote=false -a iconfont-name=font-awesome-4.4.0/css/font-awesome sample.adoc
+
+Note that `iconfont-name` does not contain the _.css_ extension and that Font Awesome local installation must contain both css and fonts folders.
 
 IMPORTANT: The default stylesheet (or any stylesheet produced by the {factory-ref}[Asciidoctor stylesheet factory]) is required for this feature to work.
 


docs/_includes/icons.adoc
```diff
@@ -61,6 +61,14 @@ Asciidoctor adds a reference to the Font Awesome stylesheet and font files serve
   href="http://cdnjs.cloudflare.com/ajax/libs/font-awesome/3.1.0/css/font-awesome.min.css">
 ----
 
+In case the stylesheet and fonts must be available offline, it is possible to set as reference a folder path relative to the document.
+
+To do so, set the attributes `iconfont-remote` to `false`, and `iconfont-name` with the path to Font Awesome stylessheet.
```
mojavelinux: Could you fix the spelling of stylesheet in this sentence?

mojavelinux:
I like it!

I'd like to make one suggestion. Could you split the attributes into two separate examples. The reason is that there is a default location where it looks for the FontAwesome stylesheet if you only specify `iconfont-remote=false`. Explain that, then explain that they can specify an explicit path using `iconfont-name`.

One of the reasons we've hesitated to document this (though I still think it's a good idea to do so), is because we may change the name of these attributes in the future to be more consistent across the board. Perhaps you can just add a reference to that discussion. See https://github.com/asciidoctor/asciidoctor/issues/659.

abelsromero:
> One of the reasons we've hesitated to document this (though I still think it's a good idea to do so)

If you thinks it's not the time to make this public, it's not a problem for me, really.
But I think it's better to document them when it is not a uncommon use case like this one. Adding the issue link and a WIP note may encourage colaboration and allows to check with the community what they think.

Let me know what do you think, really there's no need to merge the PR just to be nice ;)

mojavelinux:
I do think it's correct to document it now because there is a need for it. I just want to make sure we raise awareness that it's an ongoing discussion by linking to the issue.

abelsromero:
While reviewing the options I saw that setting only `iconfont-remote=false` sets the path to the same folder, that is `./font-awesome.css`.
But, this does not work because FontAwesome imports resources from the fonts folder with this line
```
  src: url('../fonts/fontawesome-webfont.eot?v=4.4.0');
```
This means that for this to work you must put the fonts directory in the parent directory of the document.

The simpler solution I can think off is using `'css/font-awesome'` instead of `'font-awesome'` in this line
https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/converter/html5.rb#L77.
But I am not sure, I don't understand all ruby syntax :cold_sweat: I imagine it checks for 'iconfont-name', if found it uses its value, if not it sets as path `'font-awesome'`.

Also, I saw that  iconfont-remote must be set as `iconfont-remote!!`, ` iconfont-remote=false` does not work. Is that intended?

mojavelinux:
> iconfont-remote=false does not work. Is that intended?

That's how all flags work in AsciiDoc at the moment. A false value can only be achieved by unassigning the attribute (e.g., `:iconfont-remote!:`). AsciidoctorJ adds a layer of abstraction over this behavior by recognizing a `false` boolean value, but that is not implemented in core.

mojavelinux:
> But, this does not work because FontAwesome imports resources from the fonts folder with this line

In order to use the Font Awesome stylesheet unmodified, you have to arrange the files in the way that Font Awesome expects. A simpler way to accomplish this is to use the `stylesdir` to set the location of the CSS and then put the fonts in `fonts`.

```
:iconfont-remote!:
:stylesdir: css
```

abelsromero:
Completely forgot about this one, I'll just close it

PR #417, opened 2015-10-04
remove 'caution: copycss is broken' from docs

AlanDaniels101:
This message: "The copycss attribute, which would copy a custom stylesheet when a document is rendered, is broken. Watch Issue #300 for updates." appears on the Custom Themes part of the documentation. According to https://github.com/asciidoctor/asciidoctor/issues/300, this is no longer the case.


mojavelinux:
Thanks @AlanDaniels101!

PR #425, opened 2015-11-24
updated doco for footer docinfo

noble-grant:
This is an update in relation to issue 424


mojavelinux:
Thanks @noble-grant! This revision is very well written! I especially like the emphasis you placed on replacing the built-in footer with one from the docinfo file.

Issue #430, opened 2015-12-11
[docs] Mention asciidoctor-pdf in the PDF section


jaredmorgs:
http://asciidoctor.org/docs/user-manual/#pdfs could do with a mention of the http://asciidoctor.org/docs/convert-asciidoc-to-pdf/ option, rather than just the fo version.

This content is contained in the https://github.com/asciidoctor/asciidoctor.org/blob/master/docs/user-manual.adoc#67-pdfs file.

mojavelinux:
I agree. This section was written before Asciidoctor PDF existed. But now, it's (becoming) the recommended option for PDF creation in the Asciidoctor ecosystem.

jaredmorgs:
I've got a PR I'm about to submit for this. But first, Lunch!

mojavelinux:
Nice!!

mojavelinux:
We may want to consider suggesting an HTML-to-PDF as a strategy for making a PDF. First, it's possible to simply Print to PDF, which uses the print styles that are bundled with the default stylesheet.

Another option is to use a tool that automates this process, such as WeasyPrint. The following thread on the AsciiDoc Python mailinglist describes the use of WeasyPrint.

https://groups.google.com/forum/#!msg/asciidoc/nI5g3j7eNbg/AzjkLSQ-BgAJ

I think a brief mention in the user manual might at least raise awareness that such solutions exist. We'll continue to focus on Asciidoctor PDF as our recommended path, but it's important to keep an eye on other strategies.

PR #431, opened 2015-12-11
Issue 430

jaredmorgs:
Terse improvement for #430 to add asciidoctor-pdf to the section.


mojavelinux:
Could you split off the attribute sorting at the top of the document as a separate pull request?

jaredmorgs:
I did try that but I think I munged something when I merged my branch. The
attribute sorting is in another PR.

Might need to roll back the commit.

On Fri, 11 Dec 2015, 15:59 Dan Allen <notifications@github.com> wrote:

> Could you split off the attribute sorting at the top of the document as a
> separate pull request?
>
> —
> <https://github.com/asciidoctor/asciidoctor.org/pull/431#issuecomment-163848329>
> .
>
-- 

Sent from Mobile.

mojavelinux:
Actually, the best way to do it to simply change it how you want it, then use:

```
git rebase -i origin/master
```

Change the "pick" to "s" for all lines except the first.

That will squash all the commits into a single change. Then force push to your branch.

Btw, I recommend naming branches "issue-XXX", though the way you have done it, "Issue_XXX" certainly isn't wrong.

mojavelinux:
I think there's room here to mention theming. You can mention that for fopub, it is necessary to pass XSL parameters or modify the XSLT. For Asciidoctor PDF, there is built-in theming via YAML that is documented at http://gist.asciidoctor.org/?github-asciidoctor%2Fasciidoctor-pdf%2F%2Fdocs%2Ftheming-guide.adoc.

jaredmorgs:
I think I managed to revert the change to the URI order. Before I attempt to do that again, is that something you're OK with me doing, or is the order like that for a reason.

jaredmorgs:
Anything beyond this is a bit past my GitFu to try and resolve. Next time I'll keep the changes very atomic.

mojavelinux:
Nice work!

I'm fine with you alphabetizing the attributes. They are probably the way they are due to random chance.

I do like to section the attributes using a comment line as a group header. Here's the pattern I've been using:

```asciidoc
= Asciidoctor User Manual
Author 1; Author 2
// meta:
:description: ...
:keywords: AsciiDoc
...and so on
// settings:
:compat-mode:
:doctype: book
:idprefix:
:idseparator:
:page-layout!:
:source-language: asciidoc
...and so on
// refs:
:uri-asciimath: http://docs.mathjax.org/en/latest/asciimath.html
:uri-latexmath: http://docs.mathjax.org/en/latest/tex.html
:uri-mathjax: http://www.mathjax.org
...and so on
```

This way, it's easy to find where to add a new attribute and we can split them into separate files in the future. We can define additional groups if necessary.

mojavelinux:
Here's how I define the groups

* *meta* - attributes that define information that goes into the header of the output document for indexing
* *settings* - attributes that control the conversion and formatting
* *refs* - attributes that are referenced in the content, such as URIs; prefix each attribute by its type to make auto-complete work nicer, such as `uri-` for a URI.

docs/user-manual.adoc
```diff
@@ -100,6 +100,7 @@ Sarah White <https://twitter.com/carbonfray[@carbonfray]>; Dan Allen <https://gi
 :uri-asciidoctorjs: {uri-org}/asciidoctor.js
 :uri-irc: irc://irc.freenode.org/#asciidoctor
 :uri-asciidoc-leveloffset: http://asciidoc.org/userguide.html#X90
+:asciidoctor-pdf-docs-ref: https://github.com/asciidoctor/asciidoctor-pdf/blob/master/README.adoc
```
mojavelinux: To follow the new pattern we are establishing, this attribute would be defined as:

```
:uri-asciidoctor-pdf-guide:  https://github.com/asciidoctor/asciidoctor-pdf/blob/master/README.adoc
```

I can change it after I merge. Just letting you know for future edits...and looking for your input as well.

docs/user-manual.adoc
```diff
@@ -1177,37 +1178,12 @@ include::{includedir}/manpage.adoc[]
...
-You can view the PDF using any PDF viewer.
-
-Rather than converting from AsciiDoc to DocBook and then from DocBook to PDF in two steps, +a2x+ can go directly from AsciiDoc to PDF in a single call:
-
- $> a2x -f pdf sample.adoc
-////
+fopub Tool::
+  Directions for using the tool are documented in the {fopub-doc-ref}[asciidoctor-fopub README]. To alter themes, it is necessary to pass XSL parameters or modify the XSLT.
+asciidoctor-pdf Tool::
+  Directions for using the tool are documented in the {asciidoctor-pdf-docs-ref}[asciidoctor-pdf README]. There is built-in theming via YAML, which is documented in  http://gist.asciidoctor.org/?github-asciidoctor%2Fasciidoctor-pdf%2F%2Fdocs%2Ftheming-guide.adoc[this gist].
```
mojavelinux: I think we should still mention a2x as an option, but make it last. I'll follow up with a revision.

mojavelinux:
Revision complete. Please see the result here: http://asciidoctor.org/docs/user-manual/#pdfs

Issue #432, opened 2015-12-11
[docs] Section Pending "Call To Arms"


jaredmorgs:
I think it might be a good idea to persuade readers to consider contributing to the TODO sections.

"Section Pending" suggests the content is under development, whereas it may not be. 

I'm going create a branch that links to individual tickets where we users can discuss scope of the sections and reach a collaborative decision. Tracking them in GitHub issues will keep the information flow transparent, and will allow users to submit PRs against the issue to track progress.

mojavelinux:
Great idea! This also ensures that there is a reference from the issue tracker leading back to those empty sections so that when we are doing project management on the repository we don't forget about them!

jaredmorgs:
https://github.com/asciidoctor/asciidoctor.org/pull/467 contains the fix for this.

jaredmorgs:
This issue is complete.

mojavelinux:
Excellent! Nice work.

PR #438, opened 2015-12-11
Section Pending proposed content and bug approach

jaredmorgs:
This should give you an idea about what I propose for those sections that require more info. Let me know if you like this approach and I'll continue with this PR.


jaredmorgs:
#432 relates to this PR.

mojavelinux:
I like it! Great idea!

The change for Asciidoctor PDF got mixed up in this pull request. Could you remove that part of the change and force push?

Also, it's a good idea to always squash a pull request down to a single commit, if possible. That way, the history doesn't show local changes that weren't intended as part of the effective pull request.

jaredmorgs:
I'm getting into better housekeeping for my PRs, so I'm closing this one off. I've created a PR that should be cleaner and more atomic, and will use the Amend function to push only one major change.

mojavelinux:
:+1:

PR #467, opened 2015-12-22
Issue 432 Sections Pending

jaredmorgs:
First of all, apologies for the two micro commits here instead of just one. Still getting used to working with Append, but I'll get there. 

You'll note that some Sections still are declared as Section Pending. These were parent sections with child section nodes and I felt they should not have a tracker. I can easily remedy that.


jaredmorgs:
#432 is the originating issue for this PR.

mojavelinux:
:+1:

PR #470, opened 2016-01-05
Delete "Additional Help" per Issue 464

jaredmorgs:


mojavelinux:
I recommend starting out the issue comment with the following string:

```
resolves #xyz
```

where "xyz" is the issue number. That way, when it is merged, the issue is automatically closed by the GitHub elves.

mojavelinux:
So in this case, you'd write:

```
resolves #464 delete "Additional Help" section
```

PR #473, opened 2016-01-08
Provide a tip for installing asciidoc on Windows

ltearno:
About the code page change that is required when working in windows command line

### DIFF
--- a/index.adoc
+++ b/index.adoc
@@ -116,7 +116,7 @@ TIP: The benefit of using a Linux package manager to install the gem is that it 
 The drawback is that the package may not be available immediately after the release of the gem.
 If you need the latest version, you can always fallback to using the `gem` command.
 
-TIP: If you are working on Windows, you may get this error : `ERROR:  While executing gem ... (Encoding::UndefinedConversionError)`. In that case, you have to change the code page used by your command line. This is done by executing `chcp 1252`. Then use the following commands as stated.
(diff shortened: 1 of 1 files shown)


index.adoc
```diff
@@ -116,6 +116,8 @@ TIP: The benefit of using a Linux package manager to install the gem is that it
 The drawback is that the package may not be available immediately after the release of the gem.
 If you need the latest version, you can always fallback to using the `gem` command.
 
+TIP: If you are working on Windows, you may get this error : `ERROR:  While executing gem ... (Encoding::UndefinedConversionError)`. In that case, you have to change the code page used by your command line. This is done by executing `chcp 1252`. Then use the following commands as stated.
```
mojavelinux: I like this suggestion, but Asciidoctor strongly recommends the use of UTF-8, so this should be:

```
$ chcp 65001
```

ltearno:
Tested and approved !
I updated the pull request to instruct the user to use the 65001 code page.

As I am curious, would you please kindly point to some resource explaining why 65001 and not 1252 ?

Thanks

mojavelinux:
65001 is the code for UTF-8, whereas 1252 is the code for Microsoft's extended ASCII (a blend of US-8859-1 + commonly used special symbols). UTF-8 is a far better choice. 1252 only gives you a few extra miles of gas. See https://en.wikipedia.org/wiki/Windows-1252

mojavelinux:
Thanks for adding this info!

PR #484, opened 2016-01-26
#387 documentation completed.

### LINKED ISSUE #387: Document how to split an attribute value over multiple lines
Document how to split the value of an attribute entry over multiple lines. As of 1.5.0, you end the line in a `\` to extend the value to the next line. Prior to 1.5.0, you used a `+`, though that still works too.

```
:name: first part \
second part
```

Note that the endlines are removed unless a `+` surrounded by spaces appears before the trailing backslash.

```
:poem: Roses are red, + \
Java is black.
```

jaredmorgs:
Resolves #387 by adding the following changes:

* added the primary fix of declaring how to split long value strings and preserve endlines

Additional, complimentary fixes:
* added [example] blocks to all examples and removed some lead in sentences.
* added admonition relating document attributes back to XML entities (good for migrators)
* separated what attributes can and can not do into nested lists with clear headers. Made parallel where possible.
* added bullet items relating to core issue into these restructured lists.

### DIFF
--- a/docs/_includes/attr-doc.adoc
+++ b/docs/_includes/attr-doc.adoc
@@ -38,23 +38,24 @@ The `right` value will override the default value.
 The value assigned to an attribute in the document header will replace the intrinsic value.
 
 Attribute entries are also used to store URLs.
-Here's an example:
 
+.URL attribute entry
 [source]
 ----
 :fedpkg: https://apps.fedoraproject.org/packages/asciidoc
 ----
 
 Now you can refer to this attribute entry anywhere in the document (where attribute substitution is performed) by surrounding its name in curly braces:
-Here's the attribute in use:
 
+.fedpkg attribute usage example
 ====
 [source]
 Information about the AsciiDoc package in Fedora is found at {fedpkg}.
 ====
 
-You can also set the base path to images (default: _empty_), icons (default: `./images/icons`), stylesheets (default: `./stylesheets`) and JavaScript files (default: `./javascripts`):
+You can also set the base path to images (default: _empty_), icons (default: `./images/icons`), stylesheets (default: `./stylesheets`) and JavaScript files (default: `./javascripts`).
 
+.Base path config example
 [source]
 ----
 :imagesdir: ./images
@@ -65,16 +66,26 @@ You can also set the base path to images (default: _empty_), icons (default: `./
 
 When you find yourself typing the same text repeatedly, or text that often needs to be updated, consider assigning it to a document attribute and use that instead.
 
+NOTE: If you're familiar with writing in XML, you'll recognize document attributes as (custom text) entities.
+
 Attribute entries have the following characteristics:
 
-* can be assigned to a document via the CLI or in the document's header and body
-* have default values
-** their default values are set when the user leaves the value field empty
-* some have alternate values
-* they can be unset (turned off) with a `!`
-* they override config files, intrinsic attributes
-* they do *not* override attributes issued from the command line
+Attributes entries can::
+* be assigned to a document:
+** through the CLI
+** in the document's header
+** in the document's body
+* be unset (turned off) with a `!`
+* have default values, which are set when the user leaves the value field empty
+* span multiple lines
+* override some intrinsic attributes
+* include inline AsciiDoc content.
+
+Attribute entries can not::
+
+* override attributes assigned from the command line
 ** a number of entry attributes can also be used on the command line
+* include AsciiDoc block content (for example, bulleted lists or other types of whitespace-dependent markup)
 
 ==== Attribute entry substitutions
 
@@ -111,6 +122,61 @@ Another approach is to change the order of substitutions that are applied where 
 [subs="specialcharacters,attributes,quotes,replacements,macros,post_replacements"]
 The application is called {app-name}.
 ----
+
+==== Splitting attribute entry values over multiple lines
+
+===== Splitting long value strings
+
+Use a forward-slash `\` at the end of each attribute value line to split long listings.
+
+.A long, single line attribute
+[source,asciidoc]
+----
+:long_verbose: Sometimes you have long lines of text in an attribute. You may find it more readable if you split these up neatly in the header. Doing this will help other readers understand the attribute value contents without scrolling.
+----
+
+To split the block up and make it more readable, insert a forward-slash (`\`) at the end of the line to extend the value to the next line.
+
+.A long, logically-split attribute
+[source]
+----
+:long_verbose: Sometimes you have long lines of text in an attribute.\
+You may find it more readable if you split these up neatly in the header.\
+Doing this will help other readers understand the attribute value contents without scrolling.\
+----
+
+NOTE: You can leave long lines as they are. Splitting attribute entry values such as the preceding example is only suggested for docs code readability.
+
+===== Preserving endlines
+
+Endlines are removed when the document is processed, unless a plus symbol surrounded by spaces (`&nbsp;+ \`) appears before the trailing backslash.
+
+.When endlines matter
+[source]
+----
+:haiku: Write your docs in text, + \
+AsciiDoc makes it easy, + \
+Now get back to work! + \
+----
+
+This syntax ensures that the newlines are preserved in the output document as hard line breaks.
+
+===== Limitations
+
+Attributes let you do a surprising amount of formatting for what is fundamentally a text replacement tool.
+
+It may be tempting to try and extend attributes to be used for complex replaceable markup.
+
+Supported::
+  Basic in-line AsciiDoc markup is permitted in attribute values, such as:
+* *emphasis*
+* `literal text`.
+
+Unsupported::
+  Complex AsciiDoc markup is not permitted in attribute values, such as:
+* lists
+* multiple paragraphs
+* other whitespace-dependent markup types.
 
 ////
 TODO: This section actually might make more sense in the header section.


docs/_includes/attr-doc.adoc
```diff
@@ -65,16 +68,27 @@ You can also set the base path to images (default: _empty_), icons (default: `./
...
-* they override config files, intrinsic attributes
-* they do *not* override attributes issued from the command line
+Attributes entries can::
+* be assigned to a document:
+** through the CLI
+** in the document's header
+** in the document's body
+* be unset (turned off) with a `!`
+* have default values, which are set when the user leaves the value field empty
+* have alternate values
```
mojavelinux: Let's drop "have alternate values" since I'm not sure what that really means.

docs/_includes/attr-doc.adoc
```diff
@@ -65,16 +68,27 @@ You can also set the base path to images (default: _empty_), icons (default: `./
...
+* span multiple lines
+* override config files, intrinsic attributes
```
mojavelinux: Let's change this to "override some intrinsic attributes"

since not all intrinsic attributes can be overridden and we don't have config files.

docs/_includes/attr-doc.adoc
```diff
@@ -65,16 +68,27 @@ You can also set the base path to images (default: _empty_), icons (default: `./
...
+* include inline AsciiDoc content.
+
+Attribute entries can not::
+
+* override attributes issued from the command line
```
mojavelinux: Let's change "issued" to "assigned".

docs/_includes/attr-doc.adoc
```diff
@@ -112,6 +126,65 @@ Another approach is to change the order of substitutions that are applied where
 The application is called {app-name}.
 ----
 
+==== Splitting attribute entry values over multiple lines
+
+===== Splitting long value strings
+
+Use backslashes `\` at the end of each attribute value to split long listings.
```
mojavelinux: Technically, the backslash must be preceded by a space. That space is preserved in the final value, so the line can only be split at the boundary of words (not in the middle of a word).

docs/_includes/attr-doc.adoc
```diff
@@ -112,6 +126,65 @@ Another approach is to change the order of substitutions that are applied where
...
+.A long, logically-split attribute
+[source]
+----
+:long_verbose: If you have a very long line of text \
+that you need to substitute regularly in a document, \
+you may find it easier to split it neatly in the header \
+so it remains readable to folks reading your docs code.
+----
+
+NOTE: You can leave long lines as they are. Splitting attribute entry values such as the preceding example is only suggested for docs code readability.
```
mojavelinux: Can you use sentence-per-line? Instead of period+space, use period+newline.

docs/_includes/attr-doc.adoc
```diff
@@ -112,6 +126,65 @@ Another approach is to change the order of substitutions that are applied where
...
+[example]
+.When endlines matter
+[source]
+----
+:haiku: Write your docs in text, + \
+AsciiDoc makes it easy, + \
+Now get back to work! + \
+----
+
+Preserving endlines allows you to split an attribute value across multiple lines.
```
mojavelinux: It would be more accurate to say:

This syntax ensures that the newlines are preserved in the output document as hard line breaks.

mojavelinux:
> added [example] blocks to all examples and removed some lead in sentences.

I'm not sure this is the right usage of `[example]`. Even if we decide it fits, we should almost never use `[example]` but rather the delimited block boundaries `====`. Using `[example]` is a shorthand for when you want to make an example out of a single paragraph, which is almost never the case in the user manual.

For now, I'd just leave these off this pull request.

docs/_includes/attr-doc.adoc
```diff
@@ -65,16 +68,27 @@ You can also set the base path to images (default: _empty_), icons (default: `./
 
 When you find yourself typing the same text repeatedly, or text that often needs to be updated, consider assigning it to a document attribute and use that instead.
 
+NOTE: If you're used to XML, you'll know document attributes as User Entities.
```
mojavelinux: Perhaps write this as:

If you're familiar with writing in XML, you'll recognize document attributes as (custom text) entities.

jaredmorgs: Thanks for all this feedback. Some of the comments related to the original
content that I simply reordered. Which is a prime example of surfacing
errors by improving readability.

I'll work on incorporating this feedback during my evening and we should be
good to go.

On Wed, 27 Jan 2016 at 08:15 Dan Allen <notifications@github.com> wrote:

> In docs/_includes/attr-doc.adoc
> <https://github.com/asciidoctor/asciidoctor.org/pull/484#discussion_r50909837>
> :
>
> > @@ -65,16 +68,27 @@ You can also set the base path to images (default: _empty_), icons (default: `./
> >
> >  When you find yourself typing the same text repeatedly, or text that often needs to be updated, consider assigning it to a document attribute and use that instead.
> >
> > +NOTE: If you're used to XML, you'll know document attributes as User Entities.
>
> Perhaps write this as:
>
> If you're familiar with writing in XML, you'll recognize document
> attributes as (custom text) entities.
>
> —
> <https://github.com/asciidoctor/asciidoctor.org/pull/484/files#r50909837>.
>
-- 

Sent from Mobile.

mojavelinux: > Some of the comments related to the original content that I simply reordered.

Yep, I could tell from the diff what you moved around and what you introduced. My comments pertain to the aggregate result. I appreciate that moving the content around is what helped us find the stuff that doesn't fit. So thanks!

docs/_includes/attr-doc.adoc
jaredmorgs: Incorporated.

docs/_includes/attr-doc.adoc
jaredmorgs: Incorporated.

docs/_includes/attr-doc.adoc
jaredmorgs: Incorporated.

docs/_includes/attr-doc.adoc
jaredmorgs: I'll play around with the wording of this. 

I was thinking something like:

"Use a space and backslash ` \` at the end of each attribute...

.Why is a space required?
NOTE:  The space ensures lines are only split at the boundary of words, not mid-word.

docs/_includes/attr-doc.adoc
jaredmorgs: That is both good practice and a good separation point for the sample block. Nice idea. :+1:

docs/_includes/attr-doc.adoc
jaredmorgs: Incorporated.

docs/_includes/attr-doc.adoc
jaredmorgs: All the aggregated comments above have been rolled in. I'll push the PR now.

mojavelinux:
It seems like when you do a rebase (or amend), you are getting an additional commit instead that has merge conflicts.

I'm actually very interested in your experience because if I can figure out how to explain the rebase/amend process to you, we should then be able to teach it to others.

Rather than add to this comment thread, perhaps we can exchange some e-mail about what steps you are taking, and what tools you are using, so we can get to the bottom of it. I'm not sure how good I am at explaining this, but perhaps we can work together to find the fail-proof procedure.

docs/_includes/attr-doc.adoc
mojavelinux: I'd rather it be:

"Use a single space followed by a backslash at the end of each line... The space will be preserved in the final value."

I'm not sure we want to break into an admonition here because it seems excessive and breaks the flow. I think simply stating that the space is required, and that it is preserved, allows the reader to conclude that this can only be used where there is a natural break in the text.

docs/_includes/attr-doc.adoc
mojavelinux: I'm referring to the style of writing in the user guide (not the example). Whenever you start a new sentence, begin it on a new, adjacent line instead of running sentences together on the same line.

jaredmorgs: Yep, I get this now. I'll add this standard to the contributing guide, and
will fix up this part of the PR.

I'd imagine the reason for doing this is that it makes the code more
readable in plain-text.

Should this be something we recommend to users who write AsciiDoc?

On Thu, 28 Jan 2016 at 04:44 Dan Allen <notifications@github.com> wrote:

> In docs/_includes/attr-doc.adoc
> <https://github.com/asciidoctor/asciidoctor.org/pull/484#discussion_r51029155>
> :
>
> > +:long_verbose: If you have a very long line of text that you need to substitute regularly in a document, you may find it easier to split it neatly in the header so it remains readable to the next person reading your docs code.
> > +----
> > +
> > +To split the block up and make it more readable, insert a  forward-slash (`\`) at the end of the line to extend the value to the next line.
> > +
> > +[example]
> > +.A long, logically-split attribute
> > +[source]
> > +----
> > +:long_verbose: If you have a very long line of text \
> > +that you need to substitute regularly in a document, \
> > +you may find it easier to split it neatly in the header \
> > +so it remains readable to folks reading your docs code.
> > +----
> > +
> > +NOTE: You can leave long lines as they are. Splitting attribute entry values such as the preceding example is only suggested for docs code readability.
>
> I'm referring to the style of writing in the user guide (not the example).
> Whenever you start a new sentence, begin it on a new, adjacent line instead
> of running sentences together on the same line.
>
> —
> <https://github.com/asciidoctor/asciidoctor.org/pull/484/files#r51029155>.
>
-- 

Sent from Mobile.

mojavelinux:
Merged as 5d7fc0352fde33fa0a444483c4eae06f4db155a7.

I did some additional revisions. Feel free to check my work. http://asciidoctor.org/docs/user-manual/#setting-attributes-on-a-document

I decided to take out the example blocks and just leave the caption on the listing. I don't really like how the listing block inside an example block looks...so for now I'm happy just leaving them as raw listing blocks.

jaredmorgs:
OK. No problem.

On Tue, 1 Mar 2016, 18:26 Dan Allen <notifications@github.com> wrote:

> Merged as 5d7fc03
> <https://github.com/asciidoctor/asciidoctor.org/commit/5d7fc0352fde33fa0a444483c4eae06f4db155a7>
> .
>
> I did some additional revisions. Feel free to check my work.
> http://asciidoctor.org/docs/user-manual/#setting-attributes-on-a-document
>
> I decided to take out the example blocks and just leave the caption on the
> listing. I don't really like how the listing block inside an example block
> looks...so for now I'm happy just leaving them as raw listing blocks.
>
> —
> <https://github.com/asciidoctor/asciidoctor.org/pull/484#issuecomment-190610026>
> .
>
-- 

Sent from Mobile.

PR #489, opened 2016-02-04
Add anchor definition in list items

jmini:
Discussion on the Mailing List:
http://discuss.asciidoctor.org/anchor-link-to-ordered-list-item-td4256.html

### DIFF
--- a/docs/_includes/ex-xref.adoc
+++ b/docs/_includes/ex-xref.adoc
@@ -51,6 +51,21 @@ anchor:bookmark-b[]Use a cross reference to link to this location.
 [[bookmark-c,last paragraph]]The xreflabel attribute will be used as link text in the cross-reference link.
 // end::anchor[]
 
+// tag::anchor-wrong[]
+[[anchor-point]]* list item text
+// end::anchor-wrong[]
+
+// tag::anchor-list[]
+* Fist item
+* [[step2]]Second item
+* Third item
+// end::anchor-list[]
+
+// tag::anchor-header[]
+=== [[current]]Version 4.9
+// end::anchor-header[]
+
+
 // tag::anchor-xreflabel[]
 [[tiger-image,Image of a tiger]]
 .This image represents a Bengal tiger also called the Indian tiger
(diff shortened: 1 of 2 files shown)


mojavelinux:
I like the proposal of adding a specific example (since specific examples always help bring clarity). However, I would also like to include a general description higher up that explains where anchors go in general. The description should say something like:

"In addition to being able to define anchors on sections and blocks, anchors can be defined inline where ever you can type normal text (anchors are a macro substitution). The anchors in the text get replaced with invisible anchor points in the output.

For example, you would not put an anchor in front of a list item:

```
[[anchor-point]]* list item text
```

Instead, you would put it at the very start of the text of the list item:

```
* [[anchor-point]]list item txt
```

Just keep in mind that inline anchor points are discovered where text is permitted. If text doesn't belong somewhere, then neither does an inline anchor point."

We might also want to explain the behavior of inline anchors in section titles, as they get slightly special treatment (and some caveats). But that may be for a separate issue.

mojavelinux:
cc @jaredmorgs for review

jaredmorgs:
Yeah, I like the idea of putting a more general description in as well.

@jmini would you feel comfortable amending your original PR and adding the extra info that @mojavelinux has suggested?

jmini:
Yes of course. I did not had the time to do it.

jaredmorgs:
No problem! Thanks for taking the time to do the PR.

On Mon, 8 Feb 2016, 17:35 Jérémie Bresson <notifications@github.com> wrote:

> Yes of course. I did not had the time to do it.
>
> —
> <https://github.com/asciidoctor/asciidoctor.org/pull/489#issuecomment-181240232>
> .
>
-- 

Sent from Mobile.

jmini:
I just pushed a new version of this patch.

I added an extra example of headers, derived from [this answer on StackOverflow](http://stackoverflow.com/a/35196109).

mojavelinux:
Well done! Thanks for following up @jmini and thanks for the review @jaredmorgs. Teamwork!

mojavelinux:
Did this close an issue? Just checking to make sure we didn't leave a dangling issue.

jaredmorgs:
I'm not sure there was an issue open for this @mojavelinux .

mojavelinux:
Great. Thanks for following up.

PR #490, opened 2016-02-13
Add highlightjs section in user manual - fix asciidoctor/asciidoctor#1695

andya9:


mojavelinux:
This is faaaaaaaaantastic.

I made a few notes inline to address before we go live.

docs/_includes/src-highlightjs.adoc
```diff
@@ -0,0 +1,22 @@
...
+
+[IMPORTANT]
+====
+Asciidoctor defaults to the basic bundle shipped by cdnjs, v8.9.1.
+
+To enable support for a wider range of languages and/or a newer version:
+
+. create your custom bundle in the {highlightjs-lang}[download page]
+. unpack the downloaded zip as a folder called `highlight` inside the folder of your AsciiDoc file.
+. rename `highlight.pack.js` to `highlight.min.js`.
```
mojavelinux: I think you're missing the step of setting the `highlightjsdir` attribute, right?

docs/_includes/src-highlightjs.adoc
```diff
@@ -0,0 +1,22 @@
...
+:highlightjs-lang: https://highlightjs.org/download/
+////
+
+{highlightjs-org}[HighlightJS] is a popular client-side syntax highlighter that supports a broad range of {highlightjs-lang}[languages].
+
+To use highlightjs just add `:source-highlighter: highlightjs` in your AsciiDoc file.
+
+[IMPORTANT]
+====
+Asciidoctor defaults to the basic bundle shipped by cdnjs, v8.9.1.
```
mojavelinux: Let's not mention the specific version or else we have to constantly update this when core changes. Let's just say "tracks the most recent release hosted on cdnjs". Technically, we are one behind right now, but that's because there was a breaking change in highlight.js and we haven't yet tested it.

docs/_includes/src-highlightjs.adoc
```diff
@@ -0,0 +1,22 @@
...
+- user-manual: Source Code Syntax Highlighting: HighlightJs installation
+
+:highlightjs-lang: https://highlightjs.org/download/
+////
+
+{highlightjs-org}[HighlightJS] is a popular client-side syntax highlighter that supports a broad range of {highlightjs-lang}[languages].
+
+To use highlightjs just add `:source-highlighter: highlightjs` in your AsciiDoc file.
+
+[IMPORTANT]
```
mojavelinux: I don't think this needs to be an admonition block. Let's just go with normal content.

docs/_includes/src-highlightjs.adoc
```diff
@@ -0,0 +1,22 @@
...
+Included in:
+
+- user-manual: Source Code Syntax Highlighting: HighlightJs installation
+
+:highlightjs-lang: https://highlightjs.org/download/
+////
+
+{highlightjs-org}[HighlightJS] is a popular client-side syntax highlighter that supports a broad range of {highlightjs-lang}[languages].
+
+To use highlightjs just add `:source-highlighter: highlightjs` in your AsciiDoc file.
```
mojavelinux: Core supports both "highlightjs" and "highlight.js". It's probably best to only recommend one. Which one should we recommend? (and thus stick with?)

mojavelinux: I just checked an in several other converters I only support "highlight.js", so I think we should make that the recommended setting.

andya9:
I’ll fix this and recommit :smile:

andya9:
Still, I tried it and it doesn’t work. It worked with the html footer.

mojavelinux:
Thanks for making the update!

mojavelinux:
The problem is that you also have to rename the name of the stylesheet file as well. To use the default setting, rename `github.css` to `github.min.css`.

highlight.js needs to be consistent with how it names things or we just can't map to it without modifications.

mojavelinux:
I've updated the instructions in the user manual and verified that they are correct.

mojavelinux:
This fix (and subsequent commits) addresses the following support case. https://github.com/asciidoctor/asciidoctor/issues/1695

Issue #481, opened 2016-01-22
Create a dedicated table in the user guide to document available document-level attributes


rockyallen:
Add an extra table in Appendix B so that eventually all available attributes can be found there.See also #480

mojavelinux:
I'm in favor. As I mentioned, this was one of the original objectives of the Asciidoctor user manual, so I'm happy to finally see it through!

PR #493, opened 2016-02-17
Re #481 Create table of document attributes

rockyallen:
Improve existing table first:
Added an example/notes column;
deleted doctitle because it is documented elsewhere;
added basebackend, filetype and outdir;
consistently use 'document' for the input file and 'output file' for the output.

### DIFF
--- a/docs/_includes/attr-data.adoc
+++ b/docs/_includes/attr-data.adoc
@@ -121,8 +121,8 @@ These attributes can be referenced anywhere in the document.
 
 |user-home
 |home directory of the current user.
-|If the safe mode is server or greater it resolves to ".."
-|
+If the safe mode is server or greater it resolves to ".."
+|/home/dave
 
 |===
 // end::table[]
--- a/docs/_includes/attr-data.adoc
+++ b/docs/_includes/attr-data.adoc
@@ -8,6 +8,7 @@ Added an example/notes column;
 deleted doctitle because it is documented elsewhere; 
 added basebackend, filetype and outdir;
 consistently use 'document' for the input file and 'output file' for the output.
+added safe-mode stuff
 ////
 
 Asciidoctor includes numerous attributes that are assigned values when the document is rendered.
@@ -23,12 +24,12 @@ These attributes can be referenced anywhere in the document.
 |asciidoctor 
 |Defined if the current processor is Asciidoctor.
 //|{asciidoctor}
-|(Empty string)
+|
 
 |asciidoctor-version 
 |Version of the processor.
 //|Example: {asciidoctor-version}
-|1.5.2
+|1.5.4
 
 |backend
 |Backend used to create the output file.
@@ -79,7 +80,7 @@ These attributes can be referenced anywhere in the document.
 |article
 
 |encoding 
-|Encoding of the input and output files
+|Encoding of the input and output files.
 |UTF-8
 
 |filetype 
@@ -106,6 +107,33 @@ These attributes can be referenced anywhere in the document.
 |Full path of the output directory
 //|Example: {outdir}
 |home/projects/bike/dist
+
+|safe-mode-level
+//={safe-mode-level}
+|Numeric value of safe-mode. UNSAFE is 0, SECURE is 20.
+|20
+
+|safe-mode-name
+//{safe-mode-name}
+|Text value of `safe-mode`.
+|SERVER
+
+|safe-mode-unsafe
+|Defined if `safe-mode` is UNSAFE.
+|
+
+|safe-mode-safe
+|Defined if `safe-mode` is SAFE.
+|
+
+|safe-mode-server
+|Defined if `safe-mode` is SERVER.
+|
+
+|safe-mode-secure
+|Defined if `safe-mode` is SECURE.
+|
+//={safe-mode-secure}
 
 |===
 // end::table[]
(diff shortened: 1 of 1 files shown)


mojavelinux:
This is excellent. I added a few comments inline to address.

docs/_includes/attr-data.adoc
```diff
@@ -2,58 +2,141 @@
 Included in:
 
 - user-manual: Built-in data attributes
+
+Changes:
+Added an example/notes column; 
+deleted doctitle because it is documented elsewhere; 
+added basebackend, filetype and outdir;
+consistently use 'document' for the input file and 'output file' for the output.
+added safe-mode stuff
```
mojavelinux: I don't think we need these changes in the document itself. It's sufficient to put this information in the comments of the pull request.

docs/_includes/attr-data.adoc
```diff
@@ -2,58 +2,141 @@
...
 ////
 
-Asciidoctor includes numerous intrinsic attributes that are assigned values when the document is rendered.
-The values of these built-in data attributes are derived from how the document is processed and when and where is it processed.
+Asciidoctor includes numerous attributes that are assigned values when the document is rendered.
+The values of these attributes are derived from how the document is processed and when and where is it processed.
```
mojavelinux: I propose writing this paragraph as follows:

```
Asciidoctor includes numerous attributes that get assigned values when the document is converted (i.e., intrinsic attributes).
The values of the intrinsic attributes are derived from how, when and where the document is processed (in other words, its runtime environment).
```

Note that we always want to use the term "convert" instead of "render" since render means "to display", which is the role of the browser or file viewer.

docs/_includes/attr-data.adoc
```diff
@@ -2,58 +2,141 @@
...
 These attributes can be referenced anywhere in the document.
 
 // tag::table[]
-.Built-in data attributes
-[width="70%",cols="1m,3"]
+.Document attributes - runtime information
```
mojavelinux: Let's name this:

Intrinsic document attributes related to runtime information

docs/_includes/attr-data.adoc
```diff
@@ -2,58 +2,141 @@
...
+|1.5.4
 
 |backend
-|Backend used to render document
+|Backend used to create the output file.
+//|Example: {backend}
+|html5
+
+|basebackend
+|`html` or `docbook`.
```
mojavelinux: basebackend is the backend value minus any trailing numbers (html5 -> html5, docbook5 -> docbook).

docs/_includes/attr-data.adoc
```diff
@@ -2,58 +2,141 @@
...
+|Last modified date and time of the document. [1]
+//|Example: {docdatetime}
+|2015-01-04 19:26:06 GMT+0000
 
-|docdir
-|Full path of the document directory
+|docdir 
+|Full path of the directory containing the document
+//|Example: {docdir}
+|home/projects/bike/userguide
```
mojavelinux: Let's use the following path instead:

/home/user/docs

docs/_includes/attr-data.adoc
```diff
@@ -2,58 +2,141 @@
...
 
-|docfile
-|Full path of the document file
+|docfile 
+|Full path of the document
+//|Example: {docfile}
+|home/projects/bike/userguide/master.adoc
```
mojavelinux: Let's use:

/home/user/docs/userguide.adoc

docs/_includes/attr-data.adoc
```diff
@@ -2,58 +2,141 @@
...
 
 |docname
-|Basename of the document file (no directory or file extension)
+|Name of the input document without path or extension
+//|Example: {docname}
+|master
```
mojavelinux: For consistency, make this:

userguide

docs/_includes/attr-data.adoc
```diff
@@ -2,58 +2,141 @@
...
-|doctitle
-|The title of the document
+// This isn't a document attribute, it is a header attribute and is already in <<header-summary>>
+// |doctitle
+// |The title of the document
+// |
 
 |doctype
-|Document's doctype (e.g., article)
+|Document’s doctype-article, manpage or book.
```
mojavelinux: Can you put "article, book or manpage" in parentheses?

Document type (article, book or manpage)

docs/_includes/attr-data.adoc
```diff
@@ -2,58 +2,141 @@
...
+|article
+
+|encoding 
+|Encoding of the input and output files.
+|UTF-8
```
mojavelinux: encoding is no longer defined by default (because we don't support anything except UTF-8). So we can drop this.

mojavelinux:
There are a couple of places when you write "when rendered". Could you change those to say "when converted"?

rockyallen:
Ready for review please

mojavelinux:
Merged as fd84a4620e6be00eccccb7b108ffd02bbf76a1e0.

mojavelinux:
We're still missing a few attributes in this table. In particular, there are dynamic attributes that is set per backend, basebackend, doctype and filetype.

For example, when producing HTML output with the article doctype, the following additional attributes are defined:

* doctype-article
* backend-html5
* backend-html5-doctype-article
* basebackend-html
* basebackend-html-doctype-article
* filetype-html

Perhaps you can see the pattern?

When defined, these attributes have a value of empty string.

Shall we open another issue to get these added?

PR #494, opened 2016-02-17
Document fix for https://github.com/asciidoctor/asciidoctor/issues/1140

rockyallen:
For some reason the comment in column 4 is coming out monospaced.


mojavelinux:
Thanks @rockyallen!

mojavelinux:
The monospace formatting applied to that cell is the result of known issue with complex tables. It's necessary to assign an explicit table style of none (`d`) to the cell to override the (incorrect) inherited value. I've updated the cell accordingly.

I also took the opportunity to revise that table for accuracy. Now it should be much more clear.

Issue #444, opened 2015-12-22
"Attribute conventions" Content Discussion


jaredmorgs:
http://asciidoctor.org/docs/user-manual/#attribute-conventions is currently available to create content for.

Use this issue as a tracker for content suggestions, or feel free to provide PRs referencing this ticket to keep track of contributions.

This content is contained in https://github.com/asciidoctor/asciidoctor.org/blob/master/docs/user-manual.adoc

jaredmorgs:
Perhaps describe the difference between in-document attributes and command-line attributes. I know some sections in the guide talk about command-line attributes and sometimes the difference between command-line and in-document attributes can be a little confusing to new folks.

mojavelinux:
That distinction is definitely information we want to centralize and refer
to whenever we speak of assigning attributes. It should save a lot of
redundant explanations.

mojavelinux:
We should probably mention the allowed characters for attributes (or perhaps that needs a dedicated section). The answer is that an attribute:

* must be at least one character
* the first character must be a word character (A-Z, a-z, 0-9 and _); basic latin letters only
* it can have any number of word or hyphen characters following the first character

rockyallen:
As a 'new folk' I definitely do get confused with attributes, possibly because *everything* seems to be an 'attribute'. I would like to have a go at clarifying it - perhaps starting by turning the introductory list in section 13 into a definition list. Is this a good issue to hang those PRs on?

jaredmorgs:
I seem to remember a few other issues that trusted to this as well,
@rockyallen. Including a few ideas about how to structure an attribute
reference.

See if you can find them in the issue tracker. A search for 'attribute'
should surface them.

On Thu, 18 Feb 2016, 08:54 rockyallen <notifications@github.com> wrote:

> As a 'new folk' I definitely do get confused with attributes, possibly
> because *everything* seems to be an 'attribute'. I would like to have a
> go at clarifying it - perhaps starting by turning the introductory list in
> section 13 into a definition list. Is this a good issue to hang those PRs
> on?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/asciidoctor/asciidoctor.org/issues/444#issuecomment-185446606>
> .
>
-- 

Sent from Mobile.

mojavelinux:
I'm thrilled that you are taking this on @rockyallen!

[...]

PR #496, opened 2016-02-19
Easy part of #444 - what makes a good attribute name

rockyallen:


mojavelinux:
This is excellent @rockyallen! Thanks!

Issue #480, opened 2016-01-22
Creating a dedicated table in the user guide for intrinsic attributes


rockyallen:
Add an extra table in Appendix B so that eventually all available attributes can be found there. I would also suggest renaming Appendix B to "Predefined Attributes" to make this clear.

rockyallen:
See also #481

rockyallen:
Can you confirm my interpretation of "instrinsic attribute" please:
User manual, section 12.2 "... A number of attributes and values are automatically used when the Asciidoctor processor renders a document. These attributes are called intrinsic attributes. For example, you do not need to set the normal attribute on a basic paragraph because it is set by an intrinsic attribute. But when you want a paragraph to have a different style, such as TIP, you set the TIP attribute on a paragraph and TIP will override the intrinsic normal attribute. Intrinsic attributes can be overridden or unset from the command line or within the document."

so it includes all attributes that affect the processing.
They may or may not have default values.
I don't think it includes TIP and normal, because TIP is not an attribute; it is the value of the "style" attribute, which is the first positional attribute for a paragraph.

The full list is then:
Attribute-missing
attribute-undefined
copycss
customcss
customjs
deckjs_theme
deckjs_transition
docinfo
docinfo1
docinfo2
experimental
hardbreaks
iconsdir
idprefix
imagesdir
indent
last-update-label
leveloffset
linkcss
menu
navigation
noheader
No-header-footer
notitle
nowrap
noxmlns
numbered
prewrap
role
scriptsdir
sectanchors
sectlinks
sectids
showtitle
source-highlighter
split
status
stem
style
stylesdir
stylesheet
title
toc
toclevels
toc-placement
toc-position
toc-title
version-label

Also, are "set" and "defined" synonyms?

mojavelinux:
>  "you do not need to set the normal attribute on a basic paragraph because it is set by an intrinsic attribute."

I think the user manual is creating some confusion here. I would described this example more as "intrinsic behavior" and thus a separate topic.

Intrinsic attributes are simply attributes that are defined by default that you can reference using the `{name}` syntax. These aren't to be confused with node attributes (usually block attributes), which affect the behavior of a block. The example I cited is an example of a block attribute.

mojavelinux:
> Also, are "set" and "defined" synonyms?

Yes. Feel free to make it consistent by choosing one or the other.

mojavelinux:
[...]

PR #497, opened 2016-02-19
issue #480 Create table of intrinsic attributes

rockyallen:
First go - I expect a few corrections!


jaredmorgs:
As this table gets finalised, is going to be a great addition to the guide
@rockyallen. I'll let Dan do the review on this PR as he knows his way
around the attributes.

On Sat, 20 Feb 2016, 09:58 rockyallen <notifications@github.com> wrote:

> First go - I expect a few corrections!
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/asciidoctor/asciidoctor.org/pull/497
> Commit Summary
>
>    - issue #480 Create table of intrinsic attributes
>
> File Changes
>
>    - *A* docs/_includes/intrinsic_attributes.adoc
>    <https://github.com/asciidoctor/asciidoctor.org/pull/497/files#diff-0>
>    (448)
>    - *M* docs/character-ref-attributes.adoc
>    <https://github.com/asciidoctor/asciidoctor.org/pull/497/files#diff-1>
>    (7)
>
> Patch Links:
>
>    - https://github.com/asciidoctor/asciidoctor.org/pull/497.patch
>    - https://github.com/asciidoctor/asciidoctor.org/pull/497.diff
>
> —
> <https://github.com/asciidoctor/asciidoctor.org/pull/497>.
>
-- 

Sent from Mobile.

mojavelinux:
I'll admit, it will even take me awhile to verify that we have all the attributes listed for the very reason that there is no one place that they can be found prior to now.

mojavelinux:
It's not entirely true that all intrinsic attributes have to be defined in the header. It's just that there are certain attributes that have no affect on processing *unless* they are defined in the header. The reason is, if those attributes are defined after the header, the thing that they affect has already happened.

Having said that, I think that we need a column to mark an attribute as a header attribute (i.e., an attribute that is scoped to the document header).

mojavelinux:
I think the attributes should be sorted into logical groups so that it is easier to find related attributes. Within those groups, they can still be alphabetized. But a global sorting is not helpful as it puts unrelated attributes as near neighbors.

For example:

* all caption-related attributes
* all header attributes
* etc.

We can start small and evolve the sorting over time.

mojavelinux:
Please remove any attributes related to deckjs as those should be addressed only in the documentation for the deck.js converter. Otherwise, we'd need to include the attributes for all additional converters, which would make this list unwieldy.

docs/_includes/intrinsic_attributes.adoc
```diff
@@ -0,0 +1,467 @@
...
+|Caution
+|Any
+|<<admonition>>
+
+// decided to remove this since it is not implemented and may not be an attribute even if it ever is.
+//|compact
+//|Hint to the Docbook backend to produce close spaced lists. NOT YET IMPLEMENTED.
+//|(not set)
+//|Any (value not used)
+//|-
```
mojavelinux: Yes, please remove compact.

docs/_includes/intrinsic_attributes.adoc
```diff
@@ -0,0 +1,467 @@
...
+|Replace the default stylesheet for HTML backends.
+|(not set)
+|Path
+|<<deckjs-backend-attributes>>
+
+|customjs
+|
+|(not set)
+|
+|<<deckjs-backend-attributes>>
```
mojavelinux: Remove customcss, customjs and any attribute that begins with deckjs_.

docs/_includes/intrinsic_attributes.adoc
```diff
@@ -0,0 +1,467 @@
...
+|Read input from a Docbook info file.
+|(not set)
+|Any (value not used)
+|<<docinfo-attributes-and-file-names>>
+
+|docinfo2
+|Read input from a Docbook info file.
+|(not set)
+|Any (value not used)
+|<<docinfo-attributes-and-file-names>>
```
mojavelinux: Now that we enhanced docinfo to cover all scenarios, we should drop (and otherwise stop mentioning) docinfo1 and docinfo2.

docs/_includes/intrinsic_attributes.adoc
```diff
@@ -0,0 +1,467 @@
...
+|Set the output document type.
+|article
+|article, book, inline, manpage
+|<<document-types>>
+
+|embedded
+|Is this real? related to inline doctype?
+|(not set)
+|?
+|?
```
mojavelinux: embedded is set when the header_footer option has been disabled (in other words, the document is being converted to an embeddable document). (Technically, the attribute name should be "embeddable" and may change in the future).

docs/_includes/intrinsic_attributes.adoc
```diff
@@ -0,0 +1,467 @@
...
+
+|encoding
+|The encoding of the input and output files, all backends.
+|UTF-8
+|Any valid encoding string
+|-
```
mojavelinux: encoding is no longer set by default. It can be defined by the user to control the encoding used in the output document. The value of UTF-8 will be used if the attribute is not set, but `{encoding}` will not resolve by default.

docs/_includes/intrinsic_attributes.adoc
```diff
@@ -0,0 +1,467 @@
...
+|Label for the program name liner in manpage outputs?
+|NAME
+|Any
+|?
+
+|max-include-depth
+|Not normally used. Should it be exposed?
+|64
+|-
+|-
```
mojavelinux: max-include-depth can only be set by the API, not the document. (this is a security feature)

docs/_includes/intrinsic_attributes.adoc
```diff
@@ -0,0 +1,467 @@
...
+|Used by Deck-JS backend.
+|(not set)
+|Any (value not used)
+|<<deckjs-backend-attributes>>
+
+|navigation
+|Used by Deck-JS backend.
+|(not set)
+|
+|<<deckjs-backend-attributes>>
```
mojavelinux: drop menu and navigation as these are related to deckjs.

docs/_includes/intrinsic_attributes.adoc
```diff
@@ -0,0 +1,467 @@
...
+
+|no-header-footer
+|Suppresses the rendering of the header and footer.
+|(not set)
+|Any (value not used)
+|<<doc-header>> <<footer-docinfo-files>>
```
mojavelinux: no-header-footer is not an attribute.

docs/_includes/intrinsic_attributes.adoc
```diff
@@ -0,0 +1,467 @@
...
+|Any (value not used)
+|<<doc-header>>
+
+|nofooter
+|Suppresses rendering of the footer.
+|(not set)
+|Any (value not used)
+|<<footer-docinfo-files>>
+
+|Note-caption
```
mojavelinux: Can you fix the case of "Note-caption" to be "note-caption"?

docs/_includes/intrinsic_attributes.adoc
```diff
@@ -0,0 +1,467 @@
...
+|Sets the text used to label NOTE admonitions, if icons are not used.
+|Note
+|Any
+|<<admonition>>
+
+|notitle
+|Toggles the display of a document’s title.
+|(not set)
+|Any (value not used)
+|<<document-title>>
```
mojavelinux: Using `showtitle!` is preferred. (I'll admit, there's some confusion around the relationship between these two attributes).

docs/_includes/intrinsic_attributes.adoc
```diff
@@ -0,0 +1,467 @@
...
+
+|nowrap
+|Use scrolling instead of wrapping in wide code listings.
+|(not set)
+|Any (value not used)
+|<<to-wrap-or-to-scroll>>
```
mojavelinux: nowrap is not an attribute. It's inverse, `prewrap` can be unset (using `prewrap!`) to disable wrapping.

docs/_includes/intrinsic_attributes.adoc
```diff
@@ -0,0 +1,467 @@
...
+|Registers the split module for use in the document, Deck-JS backend only.
+|(not set)
+|Any (value not used)
+|<<deckjs-backend-attributes>>
+
+|status
+|Renders the current slide number and total number of slides
+|(not set)
+|Any (value not used)
+|<<deckjs-backend-attributes>>
```
mojavelinux: Please remove split and status.

docs/_includes/intrinsic_attributes.adoc
```diff
@@ -0,0 +1,467 @@
...
+IMPORTANT,
+WARNING,
+CAUTION,
+abstract,
+partintro,
+comment,
+example,
+sidebar,
+source
+|<<style>>
```
mojavelinux: Let's drop `style` as this is not intended to be a document attribute.

docs/_includes/intrinsic_attributes.adoc
```diff
@@ -0,0 +1,467 @@
...
+|Deprecated: Use toc attribute instead.	
+|auto
+|
+|<<user-toc>>
+
+|toc-position
+|Deprecated: Use toc attribute instead.	
+|(not set = left)
+|left, right
+|<<user-toc>>
```
mojavelinux: Let's not mention `toc-placement` and `toc-position` as their use should be discouraged in favor of assigning a value to the `toc` attribute.

docs/_includes/intrinsic_attributes.adoc
```diff
@@ -0,0 +1,467 @@
...
+|Deprecated: Use custom theme instead.
+|(not set)
+|
+|[link:https://github.com/asciidoctor/asciidoctor.org/issues/379[issue #379]
+
+|untitled-label
+|?
+|Untitled
+|?
+|?
```
mojavelinux: untitled-label is the default document title when the document does not have a title.

mojavelinux:
An important attribute that is missing is compat-mode. This controls whether the processor runs in compatibility mode or the modern mode. The default is the modern mode.

mojavelinux:
Another important attribute that is missing is `allow-uri-read`. This can only be set via the API and controls whether content can be read from remote URIs.

mojavelinux:
Another attribute that is missing is `sectnums`. This controls whether sections are numbered. It is disabled by default.

mojavelinux:
The attribute `user-home` resolves to the home directory of the current user. If the safe mode is server or greater it resolves to `.`.

mojavelinux:
The attribute `asset-uri-scheme` controls which protocol is used for assets hosted on a CDN. The value values are "http", "https" or empty.

mojavelinux:
The following attributes can be set to add metadata to the HTML document:

* app-name
* description
* keywords
* copyright
* authors (which pulls from the author line)

mojavelinux:
The `iconfont-remote` attribute can be used to disable the use of a CDN for resolving the icon font (enabled by default).

mojavelinux:
The attribute `sectnumlevels` controls the depth of section numbering. The default is 3.

mojavelinux:
Let's get those changes in, then see where we stand.

Clearly, there's a bit of a fuzzy line between attributes which you can set to affect behavior and attributes that are set by default (or under certain conditions). I don't think we should try to draw this distinction. As a user, I just need to know what attributes exist, if I can set them and what values I can set them to. I can understand if some attributes have a default value and some done.

There's also a difference between a default value and a fallback value. A default value can be resolves using `{name}` by default. A fallback value is the value used internally if the attribute is not defined. You can't see fallback values by using `{name}`.

rockyallen:
Try 2:
1. Broke up "Settable attributes" into command line, header and in-document with some notes on the distinctions.
2. Grouped header attributes into running, styling, content, internationalization, and information. Quite arbitrary, would be happy to use other groupings.
Did not add user-home because it seems to fit better with the document information table.

docs/_includes/intrinsic_attributes.adoc
```diff
@@ -0,0 +1,517 @@
...
+docinfo2 = docinfo
+toc-class = use custom theme [link:https://github.com/asciidoctor/asciidoctor.org/issues/379[issue #379]
+toc-placement = toc
+notitle = showtitle!
+////
+
+NOTE: Some attributes have been deprecated or removed since Asciidoc Python and are not included here.
+See <<asciidoctor-vs-asciidoc>> if you are updating an old document.
+
+== Settable attributes
```
mojavelinux: Let's call these "secure attributes" instead.

mojavelinux: We could also go with "security attributes". To me, either one works.

docs/_includes/intrinsic_attributes.adoc
```diff
@@ -0,0 +1,517 @@
...
+|====
+
+[1] The default value isn't necesarily the value you will get by entering `\{name}`; it may be the "fallback value" which Asciidoctor uses if `name` is not defined. 
+The effect is the same either way.
+
+The opposite of header attributes, _in-document attributes_ are designed to be set in the body of the document. 
+(You could set them in the header, but then they will apply for the whole document which is not ussually what you want.)
+They are visible from the point they are set, until they are unset or assigned another value.
+
+.In-document attributes
```
mojavelinux: Perhaps we should say "Universal attributes" instead of "in-document". In mailinglist posts, I've often referred to them as "document flow attributes" (or something to that effect), meaning "in the flow of the document".

docs/_includes/intrinsic_attributes.adoc
```diff
@@ -0,0 +1,517 @@
...
+|NA
+|NA - read only
+|
+
+|====
+
+[1] The default value isn't necesarily the value you will get by entering `\{name}`; it may be the "fallback value" which Asciidoctor uses if `name` is not defined. 
+The effect is the same either way.
+
+The opposite of header attributes, _in-document attributes_ are designed to be set in the body of the document.
```
mojavelinux: I'd make it clear that it's "header or body", not just "body". The point is that they can be set either place.

mojavelinux:
I think the subgroups are a great start. They begin to bring order to the chaos. We can refine that order over time.

At the moment, the list of "header attributes" is too broad. Even though it's a good practice to define a lot of these in the header, many of them work when defined anywhere (or, at least, they work after the point they are defined).

Here's the list of attributes that I know are absolutely header attributes: (any attribute not in this list can be defined in the flow of the document)

* asset-uri-scheme
* toc
* toclevels
* toc-title
* toc-class
* sectnumlevels
* linkcss
* nolang
* lang
* encoding
* app-name
* description
* keywords
* author / authors
* email
* revnumber
* revdate
* revremark
* copyright
* stylesheet
* webfonts
* stylesdir
* icons
* iconfont-remote
* iconfont-cdn
* iconfont-name
* source-highlighter
* coderay-css
* pygments-css
* pygments-style
* max-width
* nofootnotes
* version-label
* last-update-label
* reproducible
* highlightjsdir
* highlightjs-theme
* prettifydir
* prettify-theme
* stem
* eqnums

I know I added a few more that weren't previously defined. Let's just attribute that to discovery.

There is some gray area with header attributes. For instance, `toc-title` must be defined in the header if you are using a toc with the position other than "macro". However, if you are using the macro toc, then you can define the `toc-title` anywhere before the toc is placed. The `stem` attribute must be defined in the header in order to active STEM (aka MathJax) support, but you can redefine its value later in the document to change the default dialect (asciimath or latexmath). Perhaps we can simply add that information in the comments column.

docs/character-ref-attributes.adoc
```diff
@@ -1,7 +1,12 @@
 [[char-ref-table]]
-= Built-in attributes for character replacements
+//= Built-in attributes for character replacements
+= Predefined attributes
```
mojavelinux: I kind of like the term "Built-in" rather than "Predefined". I think "predefined" are what we should call the character replacement attributes. In other words:

* built-in - reserved attribute names that affect processing in some way
* predefined - attributes that have been preset for convenience, but are otherwise just data templates

wdyt?

mojavelinux:
I think we're starting to get close. Thank you for your hard work on this, @rockyallen. This is a huge improvement.

Btw, don't worry about the CI failure. That was caused by me earlier today attempting to upgrade the software on the site. It should resolve itself once we merge.

mojavelinux:
I'd say that following should also be included in the header attributes, although they also have some gray area:

* sectids
* sectanchors
* sectlinks
* idprefix
* idseparator

The reason these are recommended to be defined in the header is that, otherwise, you don't get consistent behavior throughout the sections of the document. In other words, it's an edge case to not define these in the header.

mojavelinux:
Merged as 9940616e9f5f4ace3a2b0616fda781ff08bf16b3. See follow-up in #480.

PR #503, opened 2016-02-21
A few typos - apologies if they are actually valid American!

rockyallen:


mojavelinux:
Thanks @rockyallen! These are great catches!

Issue #499, opened 2016-02-20
Document how to use Asciidoctor with non-english languages


rockyallen:
The Asciidoc manual had Appendix D http://asciidoc.org/userguide.html#_using_asciidoc_with_non_english_languages for this topic, but the Asciidoctor manual doesn't seem to have anything similar. Should it be copied across (with suitable mods)? Where should it go? It seems too important for an annex, possibly a new section under "Enriching Your Content" or "Building Your Content". Should it also reference the ongoing discussions about translation? This section would also be good place to list the attributes like warning-label that are not otherwise documented.

mojavelinux:
I definitely support the idea of adding a section about using Asciidoctor with other languages. This topic was not yet covered because I was waiting on the completion of https://github.com/asciidoctor/asciidoctor/issues/1129. However, that's not a good reason to keep the information out of the user guide, so I think we should proceed with how it works today and adjust it once we roll out that feature.

mojavelinux:
Have you sent a pull request for this yet? I can't seem to find it.

PR #504, opened 2016-02-23
Resolves issue #499 Document use of asciidoctor with non-english lang…

rockyallen:
…uages


mojavelinux:
The changes for this are now part of #505. Closing at the request of @rockyallen.

PR #508, opened 2016-02-25
Resolves issue 495 literals

rockyallen:
Ready for review. 
Feel free to delete the footnotes-I added them to fix my confusion, but I am not sure they add that much for anyone else. 
I tried putting the attributes in alphabetic order, but thinking about it, people are more likely to scan for the character they want to insert then read the attribute name. This is easier with the previous groupings.


mojavelinux:
Nice job!

I think I might change the first footnote to an admonition, but I'll do that in a post-merge step.

PR #510, opened 2016-02-25
Issue480try2

rockyallen:
Somehow I cocked  up the previous PR for 480. Hopefully this one will work and the previous can be deleted.


rockyallen:
Please review. Main changes:
1. Merge header and document tables again, but add column and notes to differentiate.
2. Add new attributes from comments.
3. Remove embeddable and compatibility because they appear to be runtime information so should be in that table.
4. Add new group 'metadata' inspired by the sort of things that would go in a docbook info file.

mojavelinux:
This is excellent. I'm going to go ahead and do a rebase to get this into master, then do a bit of refinement after that (that should save the trouble of more merge nightmares).

One thing I'd like to change is that I'm going to switch the "Where" column to "Header-only" and use a checkmark to indicate the state. That's because this is just a binary field (either required in header or not). We'll see how that feels. If it still doesn't sit right, we can think about another way to express it.

mojavelinux:
Merged as 9940616e9f5f4ace3a2b0616fda781ff08bf16b3. Thanks!!

Issue #506, opened 2016-02-24
Include directive


rockyallen:
Section is an excellent reference, in that it gives all the information you could possibly need, but:

* For an essential topic, it is introduced quite late in the guide (section 50)
* It dives straight into the detail without demonstrating how it 'just works' for the most common use case of a document divided into chapter files
* It references the Asciidoc manual

Can I suggest a minor reorganization to remove the link, give a gentler introduction, and cover the common use cases first?

mojavelinux:
Absolutely!

It seems like it would fit well under the "Building Your Content" part, since it directly relates to the building/assembly aspect. Perhaps after (or before) "Cross references". The pending section on Preprocessor conditionals should be a direct neighbor of this section since they go hand-in-hand. (See #283).

I think that the first subsection of "Include directive" should be "Partitioning chapters", as you have hinted. There's a partial example at the top of the section, but it's not explicit and lacks detail.

For context, the section you are referring to is http://asciidoctor.org/docs/user-manual/#include-directive.

mojavelinux:
(In general, the sections in Building Your Content" could probably stand to be rearranged.)

mojavelinux:
I'm really happy with the improvements to the Include Directive section. I think the information is much clearer and more complete now. If you see room for further refinements, don't hesitate to propose them!

rockyallen:
Thanks. And I like your refinements - reads much more professionally now!

mojavelinux:
I couldn't have done it without you giving me such a solid starting point. Teamwork!

PR #511, opened 2016-02-26
Re issue #506 includes

rockyallen:
For review please.
The only problem that I know of is that the footnote link goes to the wrong place, and I am not sure why.

Slightly worried that I have gone from too "reference" to too "tutorial"!

### DIFF
--- a/docs/_includes/include-directive.adoc
+++ b/docs/_includes/include-directive.adoc
@@ -1,60 +1,186 @@
-////
-== Include Directive
+//== Include directive
 
-- User manual
-////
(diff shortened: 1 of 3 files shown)


mojavelinux:
I like it!

I'm going to merge and then follow-up with some revisions. I'll let you know when I'm done so you can review.

mojavelinux:
The section is now ready for review. See http://asciidoctor.org/docs/user-manual/#include-directive

I determined that the information you had in the footnote was important enough to include as part of the main content. The problem with the footnote jumping to the wrong location is likely a bug in Asciidoctor that appears with large documents. It may already be fixed in master because I remember addressing it recently.

PR #512, opened 2016-02-26
Recommend using `adoc` as file extension

fwilhe:
I think we should make this clear and I see `adoc` as the clear winner.


mojavelinux:
I agree! Thanks!

PR #524, opened 2016-03-01
Rework Text on Section Titles

fwilhe:


fwilhe:
I think we should shorten this section. The mapping-table for Setext should in my opinion be removed. Does this really provide value to the reader? The recommendation is to use Atx anyway. Also, this document is not a asciidoc cheat sheet or syntax reference.

Anyone disagrees?

fwilhe:
I added a separate commit for removing the table. I kept the information by writing a sentence. I think this is more appropriate for its purpose.

mojavelinux:
I agree.

I think we should take this even a step further. I think we should stop recommending the Setext-style for the document title. The main reason is, that puts Asciidoctor into compatibility mode by default. While it's possible to switch compatibility mode back off, using Setext is now more than just a style change.

I think we should mention Setext only to acknowledge its existence for historical purposes, then instruct readers not to use it.

fwilhe:
> I think we should mention Setext only to acknowledge its existence for historical purposes, then instruct readers not to use it.

Full ack.

PR #525, opened 2016-03-04
Update asciidoctor-diagram.adoc

robinbolton:
Removed mention of --pre flag requirement during installation as it appears to be no longer relevant.


mojavelinux:
Thanks Robin!

PR #528, opened 2016-03-05
Resolves #521, move social links to footer

### LINKED ISSUE #521: Move social links to the footer
I think we should move "social" links to the footer:

footer

Mogztter:
Also replaced IRC by Gitter (as suggested by @mojavelinux)

Resolves #521

### DIFF
--- a/_layouts/base.html.slim
+++ b/_layouts/base.html.slim
@@ -115,42 +115,6 @@ body.antialiased
             br
             br
             |In addition to an AsciiDoc processor and a collection of stylesheets, the project provides plugins for Maven, Gradle and Guard and packages for operating systems such as Fedora, Debian and Ubuntu. It also pushes AsciiDoc to evolve by introducing new ideas and innovation and helps promote AsciiDoc through education and advocacy.
-          p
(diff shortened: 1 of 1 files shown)


mojavelinux:
Merged as f5eaa5dc071e257b398e95f2bed8a6bb0dc6ed8b with follow-up revisions made in a66f4086548350ad051cdefa53d0830b4a497b60.

Let me know how you like the new layout!

mojavelinux:
In additional to some layout changes, I also did the following:

* switched to SVG badge (using shields.io)
* assigned a fixed width to icons by adding the fa-fw class

Mogztter:
I like it :+1: 

We also need to update the index page to remove IRC and maybe use the badge for Gitter in the footer ?
 
index

mojavelinux:
Keep in mind that any change to the index page right now is really a change to the Asciidoctor README.

I agree it's probably time to drop (or at least comment out) the IRC since most of us (none of us?) are there.

I think the Gitter badge in the footer would be overkill design-wise.

mojavelinux:
Update made.

PR #529, opened 2016-03-05
Adds a "toggleable" navigation pane (TOC) on documentation page

Mogztter:
I've added an Awestruct's extension to register Asciidoctor::Document into Awestruct's page context.
Here's a preview of the rendering:

asciidoctor_toc_pipeline

If we decide to remove the right side bar, TOC could be on the left.


_ext/asciidoctor_toc.rb
```diff
@@ -0,0 +1,18 @@
+require 'asciidoctor'
+
+module Awestruct::Extensions
+  class AsciidoctorTOC
+    def execute(site)
+      site.pages.each do |page|
+        if page.source_path.end_with? '.adoc'
+          document = Asciidoctor.load_file page.source_path, header_footer: true, safe: :safe, attributes: 'idprefix idseparator=-'
+          page.toc = true
```
Mogztter: Maybe we could go further and store Asciidoctor:Document in page context ? https://github.com/awestruct/awestruct/issues/288

_layouts/base.html.slim
```diff
@@ -93,6 +93,12 @@ body.antialiased
       - unless page.url.start_with? '/news/'
         .large-4.columns.secondary
           - if page.url =~ /^\/docs\/.+/
+            - if page.toc
```
Mogztter: @mojavelinux Is there an easy way to generate TOC (HTML) from an Asciidoctor::Document ?

mojavelinux: The correct way to generate a toc is to delegate to the outline node handler. You can see an example here:

https://github.com/asciidoctor/asciidoctor-backends/blob/master/slim/html5/block_toc.html.slim

mojavelinux: "outline" (or, formally, "block_outline") is a pseudo node (meaning it only exists by intent during conversion).

_ext/asciidoctor_toc.rb
mojavelinux: Yes, I think the right solution is to put the Asciidoctor::Document into the page context (perhaps enabled globally via an option in the site configuration). I'd still be open to using an extension that provides a simpler API to generate the toc (hiding the call to the outline handler), but we shouldn't have to parse the document a second time. That would be nice to avoid.

_layouts/base.html.slim
mojavelinux: We deprecated `convert_with_options` in favor of a consolidated `convert` method, so that call should be:

```
document.converter.convert document, 'outline', toclevels: 3
```

_ext/asciidoctor_toc.rb
mojavelinux: It's just a one-line change. If you'd like, you can send the PR for it to Awestruct.

In the meantime, it's possible to monkeypatch the AsciidoctorTemplate in _ext/pipeline.rb to modify the evaluate method before the fix is available.

```ruby
require 'awestruct/handlers/template/asciidoc'

class Awestruct::Tilt::AsciidoctorTemplate
  def evaluate(scope, locals)
    @output ||= (scope.document = ::Asciidoctor.load(data, options)).convert
  end
end
```

Mogztter:
Toggle ON
navigation-enabled

Toggle OFF
navigation-disabled

Mogztter:
@mojavelinux I've decided to create a dedicated layout for documentation page (I find it easier to understand)

mojavelinux:
> I've decided to create a dedicated layout for documentation page

:+1:  The only reason we don't have more layouts is because I was procrastinating.

mojavelinux:
I love the navigation toggle! This looks really nice.

My only critique is that we get the buttons to fit within the with of the sidebar. Perhaps we can shorten "Navigation" to "TOC"? I think that's a well understood term, especially in the Asciidoctor world.

Mogztter:
toolbar-width

I've reduced the size of the buttons and shorten "Navigation" to "Nav" instead of "TOC" to avoid the repetition with "Table Of Contents" just below.

_ext/pipeline.rb
```diff
@@ -1,6 +1,7 @@
 require 'compass'
 require 'zurb-foundation'
 require 'awestruct_ext'
+require 'asciidoctor_ext'
```
mojavelinux: I think we should put this patch in awestruct_ext instead of asciidoctor_ext. No need for the extra file right now.

_layouts/documentation.html.slim
```diff
@@ -0,0 +1,33 @@
+---
```
mojavelinux: Let's call this file _layouts/docs.html.slim. That way, it matches the folder name.

docs/user-manual.adoc
```diff
@@ -6,6 +6,7 @@ Sarah White <https://twitter.com/carbonfray[@carbonfray]>; Dan Allen <https://gi
 :keywords: AsciiDoc, Asciidoctor, syntax, reference, learn, how to, AsciiMath, LaTeX math, MathJax, Pygments, CodeRay
 :doctype: book
 :compat-mode:
+:awestruct-layout: base
```
mojavelinux: Drop the awestruct-layout attribute here.

mojavelinux:
That looks good!

I added a few inline comments. I think we're getting very close.

Mogztter:
Done!
I've updated slightly the style. In my opinion:

 * G+ button should be in the footer (not on the menu)
 * "Get writing" button color doesn't fit with the overall design
 * Black is too dark on the menu

### Blue
update-menu

I think we should align blue color (I love the header background color so maybe we should use this color everywhere we use blue ?)

### Green
update-style

### Red
red

mojavelinux:
@graphitefriction Any preference/thoughts on the color scheme? We're not looking for sweeping changes here, just something to that will get us more consistent.

mojavelinux:
In fact, let's just get rid of the +1 button. That was a failed experiment by Google and its time has passed.

Mogztter:
> In fact, let's just get rid of the +1 button. That was a failed experiment by Google and its time has passed.

I've created a dedicated issue: #543 

> I think we should move away from the yellow in favor of something more in the amber range. The yellow looks like a Volvo from the 80s.

Done, I've replaced yellow with amber.

> I think I like the idea of amber with a black font. That inverts the contrast from the menu items, setting if off nicely. This is an idea inspired by Material Design.

I've tried amber with a black font but the contrast is not great. To clarify you want to change the background color of the meny from black to amber and change font color from white to black ?

I didn't touch the background color of the menu because I don't know how to override a color in Foundation (with `_settings.scss`). I think we should open a dedicated issue to improve color consistency and minimize styles changes in this pull request. Wdyt ?

mojavelinux:
> I've tried amber with a black font but the contrast is not great. To clarify you want to change the background color of the meny from black to amber and change font color from white to black ?

I'm specifically talking about the "Get Writing!" button. I was thinking something like:

get-writing

That's close to your "Green" screenshot, except using more of an amber / goldenrod color.

The background of the navigation bar is still black (or dark gray) and the navigation links still white.

mojavelinux:
>  I think we should open a dedicated issue to improve color consistency and minimize styles changes in this pull request. 

:+1:

mojavelinux:
The topbar in Foundation is one of the hardest things to change (personally, I don't think they got the variables right). You'd start with:

```
$topbar-bg: <color>;
```

However, there are a lot of other variables involved. You can see them all here:

https://github.com/zurb/foundation-sites/blob/v4.3.2/scss/foundation/_variables.scss#L1256-L1325

Mogztter:
> I'm specifically talking about the "Get Writing!" button. I was thinking something like:
> That's close to your "Green" screenshot, except using more of an amber / goldenrod color.
> The background of the navigation bar is still black (or dark gray) and the navigation links still white.

Ok!
I've used "amber accent-4" from http://materializecss.com/color.html
amber-menu

mojavelinux:
Look good!

I'll do a final in-depth code review tonight and send a PR to your branch if I see any way to improve on it. Then, we'll be on the home stretch to merge!

Mogztter:
\o/

mojavelinux:
I sent a code review PR.

There are still a few outstanding issues to address:

* margin above and below document control bar still seems a bit off
* GitHub icon size/alignment is slightly off vertical center (perhaps we could just remove it)
* spacing below document control bar needs to be fixed on mobile; probably don't need the horizontal rule

I wasn't a fan of the background color on hover for the TOC. For now, I think the TOC should be more muted. I think I like the background color better than the border.

Mogztter:
> I wasn't a fan of the background color on hover for the TOC. For now, I think the TOC should be more muted. I think I like the background color better than the border.

Taste and colours... I like the background color on hover because you can see a response from the UI (inspiration: http://groovy-lang.org/groovy-dev-kit.html). But I'm fine with a more muted TOC.

> There are still a few outstanding issues to address:
> * margin above and below document control bar still seems a bit off
> * GitHub icon size/alignment is slightly off vertical center (perhaps we could just remove it)
spacing below document control bar needs to be fixed on mobile; probably don't need the horizontal rule

I don't like the style of the buttons on the document control bar. The buttons look like concrete (material) as opposed to the "light" style of the TOC. Maybe we should just use links within a bar with a background color (ie. an horizontal menu with items) ?
Anyway I think we should move forward with this pull request and iterate with design improvements.

mojavelinux:
> Anyway I think we should move forward with this pull request and iterate with design improvements.

Agree. The door is never closed on design :)

> I like the background color on hover because you can see a response from the UI (inspiration: http://groovy-lang.org/groovy-dev-kit.html).

Giving it some more thought, it wasn't really the background color that bothered. It was the fact that we needed to add extra padding on the link to make it look right, and there is already so little space. With a bit more work, there's a way to accomplish this without losing horizontal space. So let's focus on that change in a follow-up.

mojavelinux:
There is a VIM swap file in this PR. Can you remove it?

mojavelinux:
Actually, I'll take care of it when I merge.

mojavelinux:
Merged as f620320dbbe53ea6da778ae97d14a18839004540 with a few follow-up tweaks in 8f9252853ed5f4bf57869e56ce92f2a39f75bdfc.

mojavelinux:
Nice work, @Mogztter!

Mogztter:
I'm late to the party, sorry about the swap file :neutral_face: 
And it's already live, cool!

PR #531, opened 2016-03-06
resolves #526 upgrade to Awestruct 0.5.7

### LINKED ISSUE #526: Upgrade to Awestruct 0.5.7
Upgrade to Awestruct 0.5.7. This upgrade will bring a smarter development mode that can be used with LiveReload.

The one hangup right now for upgrading is that the Atom feed doesn't generate for some reason.

This will force a Sass upgrade, so we'll also need to change the Sass rule in `.exampleblock > .content` from:

```
@include single-box-shadow($panel-border-color, 0, 1px, 4px);
```

to

```
@include single-box-shadow(0, 1px, 4px, null, $panel-border-color);
```

mojavelinux:


Rakefile
```diff
@@ -91,7 +95,7 @@ desc 'Generate and preview the site locally using the specified profile (default
 task :preview, [:profile] => :check do |task, args|
   profile = args[:profile] || 'development'
   profile = 'production' if profile == 'prod'
-  run_awestruct %(-P #{profile} -g -s)
+  run_awestruct %(-P #{profile} -a --generate-on-access --livereload -s)
```
Mogztter: livereload :heart:

mojavelinux: It's pretty sweet, indeed.

Mogztter:
LGTM, just tried it locally.
Live reload is not triggering a refresh on my browser (Chrome). Do I need to setup something special ?
Otherwise I can see my changes if I manually refresh my browser without restarting Awestruct.

mojavelinux:
I think you have to activate live reload manually the first time (at least, I had to do that step). After that, it seems to stay enabled for the same base URL. For some reason, Awestruct doesn't auto-activate it like other applications do.

mojavelinux:
I recommend reporting the auto-activate thing upstream in Awestruct.

Issue #434, opened 2015-12-11
"Migrating from AsciiDoc Python" Content Discussion


jaredmorgs:
http://asciidoctor.org/docs/user-manual/#compared-to-asciidoc is currently available to create content for.

Use this issue as a tracker for content suggestions, or feel free to provide PRs referencing this ticket to keep track of contributions.

elextr:
I think you are doing asciidoctor a disservice with this section, it is not competing with asciidoc, it does not need to be compared with it, there is a detailed annex for those who need to know the specifics for porting documents.

Just acknowledge Asciidoc Python as the initial implementation that compatibility mode matches and thats all that is necessary.

mojavelinux:
I agree with @elextr. AsciiDoc is the language, Asciidoctor is the processor (just as AsciiDoc Python is a processor).

This section made it's way into the document from an early outline, but I don't think there's a need for it. The correct place for any content discussing the differences between Asciidoctor and AsciiDoc is the compatibility section (Appendix A, http://asciidoctor.org/docs/user-manual/#asciidoctor-vs-asciidoc). That section could perhaps have a small migration guide for people switching from the AsciiDoc Python processor to the Asciidoctor processor.

jaredmorgs:
Feel free to do a PR to delete this section, @elextr

Unless I beat you to it. 😉

On Sun, 17 Jan 2016, 12:03 Dan Allen <notifications@github.com> wrote:

> I agree with @elextr <https://github.com/elextr>. AsciiDoc is the
> language, Asciidoctor is the processor (just as AsciiDoc Python is a
> processor).
>
> This section made it's way into the document from an early outline, but I
> don't think there's a need for it. The correct place for any content
> discussing the differences between Asciidoctor and AsciiDoc is the
> compatibility section (Appendix A,
> http://asciidoctor.org/docs/user-manual/#asciidoctor-vs-asciidoc). That
> section could perhaps have a small migration guide for people switching
> from the AsciiDoc Python processor to the Asciidoctor processor.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/asciidoctor/asciidoctor.org/issues/434#issuecomment-172284412>
> .
>
-- 

Sent from Mobile.

rockyallen:
[...]

PR #532, opened 2016-03-06
Re issue #434 migration guide.

rockyallen:
Some of it is guess work based on the docs. Please confirm my interpretation.

### DIFF
--- a/docs/_includes/attr-builtin.adoc
+++ b/docs/_includes/attr-builtin.adoc
@@ -570,6 +570,13 @@ Mutually exclusive with the `notitle` attribute.
 |{y}
 |<<stem-in>>
 
+|tabsize
+|Converts tabs to spaces in verbatim content blocks (eg listing, literal).
+|4
+|1 or more
+|
+|-
+
 |webfonts
 |Control whether webfonts are loaded when using the default stylesheet.
 |_empty_
--- a/docs/migrating-from-asciidoc.adoc
+++ b/docs/migrating-from-asciidoc.adoc
@@ -1,0 +1,341 @@
+// Issue #480. #434, #463 refer.
+:ap: Asciidoc Python
+:ad: Asciidoctor
+:url-diagram: https://github.com/asciidoc/asciidoc-diagram
+:url-migrate: https://github.com/asciidoc/docs/migration
+:url-recommended: http://asciidoctor.org/docs/asciidoc-recommended-practices
+:url-diffs: http://asciidoctor.org/docs/asciidoc-asciidoctor-diffs/
+== Migrating from {ap} to {ad}
+
+The purpose of this section is to get documents that converted correctly using {ap} to produce the same output using {ad}.
+The differences are very small so that most documents will need very few or no changes.
+Once you have done the basic changes, this section also describes what to change to make use of all the new features in {ad}.
+
+=== Command line
+
+{ad} was designed from the outset to be a drop-in replacement for {ap}. 
+For most documents therefore you can just replace the {ap} call:
+
+`{ad} mydocument.adoc`
+
+with this {ad} call:
+
+`{ad} -a compat-mode mydocument.adoc`
+
+If this doesn't work for you, check out the differences below, particularly <<migrate-deprecated>> and <<migrate-extensions>>.
+
+==== internationalisation
+
+// ref ap UG, not checked
+{ap} has built-in .conf files so that it can translate captions just by setting `-a lang=<language code>`. 
+See <<language-support>> for the {ad} equivalent.
+
+==== Configuration files
+
+{ad} does not use .conf files or filters, so, `--conf-file`, `-dump-conf`, and `--filter` are not implemented. 
+
+==== Themes
+
+// ref ap UG. Actually seems to set JS as well. OK to ignore?
+{ap} themes are just stylesheets, so replace `--theme` with `-a stylesheet=<stylesheet>`.
+
+==== Compact
+
+// ref ap UG. 
+{ap} `--compact` reduces line spacing in lists. 
+To achieve the same in {ad}, use a custom stylesheet.
+
+==== Doctest
+
+// ref ap UG. 
+{ap} `--doctest` runs units tests. 
+See GitHub for how to run the {ad} unit tests.
+
+==== Default html backend
+
+{ap}: xhtml (version 1.1).
+ad}: html5. 
+Use `--backend xhtml` to replicate {ap} behaviour.
+
+=== Changed syntax
+
+// Simplified from migration guide
+These changes are non backward-compatible, but if you set the the `compat-mode` attribute, {ad} will accept the {ap} syntax.
+For the long term you should convert to the {ad} syntax.
+
+[cols="1a,1a,1a"]
+|====
+|Feature |{ap} or {ad} with `compat-mode` |{ad}
+
+|_italic text_
+|pass:['italic text']
+|pass:[_italic text_]
+
+|`monospaced text`
+|pass:[+monospaced text+]
+|pass:[`monospaced text`]
+
+// Check - I may be confused
+|Prevent attribute replacement
+|pass:[`{asciidoc-version}`]
+|pass:[`+{asciidoc-version}+`]
+
+|"`double quotes`"
+|pass:[``double quotes'']
+|pass:["`double quotes`"] or enter them directly
+
+|'`single quotes`'
+|pass:[`single quotes']
+|pass:['`single quotes`'] or enter them directly
+
+|Document title
+|
+[source]
+----
+Title
+=====
+----
+|
+[source]
+----
+= Title
+----
+
+|====
+
+// I am sure that this was good idea at the time, but...
+CAUTION: Do NOT use the underline-with-equals form for the document title; this implicitly sets `compat-mode`.
+
+For full details, see the {url-migrate}[Migration Guide].
+
+These non backward-compatible changes are not affected by the `compat-mode` attribute:
+
+[cols="1a,1a,2a"]
+|====
+|Feature |{ap} |{ad}
+
+|Index terms
+|pass:[`++`] and pass:[`+++`]
+|pass:[((Sword))] and pass:[(((Sword, Broadsword, Excalibur)))]
+
+|Underlined titles
+|Flexible
+|Underline length must match title length +/- 1 character (Underlined titles are deprecated anyway)
+
+|+ifeval::[]+ 
+|Flexible
+|Compares the values of attributes only
+
+|Block delimiters
+|Flexible
+|The length of start and end delimiters for a block must match
+
+|Asciidoc cell formatting (e.g `a\|`)
+|`a` or `asciidoc`
+|`a` only
+
+|Deprecated tables
+|Yes
+|No (you don't want them anyway)
+
+|====
+
+
+[[migrate-deprecated]]
+=== Deleted and deprecated attributes
+
+// Based on the previous table, but a lot of things shown as not-implemented actually are.
+These are attributes that either no longer exist, work differently, or have better alternatives.
+
+[cols="<20,<20,<60"]
+|====
+|{ap} |{ad} |Notes
+
+
+|`big`, `small`, `underline`, `overline`, `line-through`, colors
+|_deprecated_
+|Character attributes to apply formatting directly. 
+Usually better to apply a role, then apply the formatting based on that role by using a stylesheet.
+
+// Checked
+|`indoc`
+.2+|_not implemented_
+.2+|Provides the name and directory of the current document. 
+(Distinct from `docname`, because `indoc` may be an included document, and `docname` is always the master document).
+No {ad} equivalent.
+
+// Checked
+|`indir`
+
+|`asciidoc`
+|`asciidoctor` 
+|{ap} sets `asciidoc` to show that it is the current processor. 
+{ad} sets `asciidoctor` instead.
+
+// Ref migration guide
+|`toc2`
+.2+|`toc`
+.2+|Combined in a single attribute, see <<toc>>.
+
+// Ref migration guide
+|`toc-placement`
+
+// Ref ap UG. I know it workls in ad tables; no ref to ad paragrphs so assume not implemented.
+|options="unbreakable"
+|
+|In {ad}, this only works for tables, not paragraphs.
+
+// No checked
+|options="pgwide"
+|_not implemented_
+|DocBook attribute to make tables full page width, whatever the current indent.
+No {ad} equivalent.
+
+// old table
+|`replacements2`
+|`post_replacements`
+|Renamed.
+
+// http://www.methods.co.nz/asciidoc/userguide.html#X39
+// not sure if a source block is exactly the same thing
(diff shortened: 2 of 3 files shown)


docs/migrating-from-asciidoc.adoc
```diff
@@ -0,0 +1,341 @@
...
+==== Themes
+
+// ref ap UG. Actually seems to set JS as well. OK to ignore?
+{ap} themes are just stylesheets, so replace `--theme` with `-a stylesheet=<stylesheet>`.
+
+==== Compact
+
+// ref ap UG. 
+{ap} `--compact` reduces line spacing in lists. 
+To achieve the same in {ad}, use a custom stylesheet.
```
mojavelinux: AsciiDoc Python never had a compact option. Asciidoctor briefly had a compact option, but it was for compressing whitespace in the generated file. Can you drop this section?

docs/migrating-from-asciidoc.adoc
```diff
@@ -0,0 +1,341 @@
...
+
+// ref ap UG. 
+{ap} `--doctest` runs units tests. 
+See GitHub for how to run the {ad} unit tests.
+
+==== Default html backend
+
+{ap}: xhtml (version 1.1).
+ad}: html5. 
+Use `--backend xhtml` to replicate {ap} behaviour.
```
mojavelinux: There is no way to exactly replicate the AsciiDoc Python behavior here. Our xhtml is not the same as xhtml from AsciiDoc Python. The xhtml 1.1 backend was for a web of a different era and that web is dead. It's the web that moved forward. I think it's okay to say that the default backend mostly closely matches the html5 backend in AsciiDoc Python and leave it at that.

docs/migrating-from-asciidoc.adoc
```diff
@@ -0,0 +1,341 @@
...
+|
+[source]
+----
+= Title
+----
+
+|====
+
+// I am sure that this was good idea at the time, but...
+CAUTION: Do NOT use the underline-with-equals form for the document title; this implicitly sets `compat-mode`.
```
mojavelinux: I don't think we should say "do NOT use". Instead, we should say "if you use it, then it implicitly sets `compat-mode`. The correct name for this form is "Setext-style" (not underline-with-equals).

docs/migrating-from-asciidoc.adoc
```diff
@@ -0,0 +1,341 @@
...
+// old table
+|`replacements2`
+|`post_replacements`
+|Renamed.
+
+// http://www.methods.co.nz/asciidoc/userguide.html#X39
+// not sure if a source block is exactly the same thing
+|`plaintext`
+|_not implemented_
+|Use a source block.
```
mojavelinux: In my opinion, plaintext was a huge hack that should have never been added to AsciiDoc Python. It disables substitutions globally, which can really wreck the whole behavior of the parser.

I think the closest equivalent is to wrap the content in a pass block.

docs/migrating-from-asciidoc.adoc
```diff
@@ -0,0 +1,341 @@
...
+
+// Not checked
+|specialwords
+|_not implemented_
```
mojavelinux: We should say "can be implemented using an extension" (probably a Postprocessor).

docs/migrating-from-asciidoc.adoc
```diff
@@ -0,0 +1,341 @@
...
+In {ad} use a role and stylesheet, or a user-defined attribute, or define a macro.
+
+// does this need an explanation?
+|presubs
+|-
+|Not required
+
+// old table says not, but...
+|halign, valign for table cells
+|Use new table notation `^.^\|` etc
```
mojavelinux: The correct way to describe this is "use column or cell specifiers instead"

docs/migrating-from-asciidoc.adoc
```diff
@@ -0,0 +1,341 @@
...
+|See <<cell>>.
+
+|sgml
+|_not implemented_
+|No {ad} equivalent (SGML is so last century).
```
mojavelinux: I don't even think "sgml" is worth mentioning. It's probably used by, well, nobody.

docs/migrating-from-asciidoc.adoc
```diff
@@ -0,0 +1,341 @@
...
+
+// Actually I would like this. Did I see a sample extension for it?
+|showcomments
+|_not implemented_
+|Turn single line comments into DocBook "remark" elements.
+No {ad} equvalent.
+
+|tabsize
```
mojavelinux: Asciidoctor supports tabsize, but only for blocks with a verbatim content model (listing, literal, etc)

docs/migrating-from-asciidoc.adoc
```diff
@@ -0,0 +1,341 @@
...
+[[migrate-stylesheet]]
+=== Default HTML stylesheet
+The {ap} and Acidoctor stylesheets are completely compatible, because their formatting is based on the same role attributes.
+If you happen to prefer the {ap} stylesheet, you can use it by copying it from the {ap} "stylesheets" directory and telling {ad} to use it with `-a stylesheet=asciidoc.css`.
+
+// I may be doing something wrong. Can't believe no-one else has had this problem!
+
+IMPORTANT: The {ad} stylesheet (unlike the {ap} stylesheet) loads its font based icons from the web, therefore does not work well offline. To solve this, do not use inline icons <<inline-icons>>.
+
+By default, {ad} links to the default stylesheet (e.g., +linkcss+), but {ap} embeds it.
```
mojavelinux: Asciidoctor does embed the default stylesheet by default, just like AsciiDoc Python. The behavior is the same.

docs/migrating-from-asciidoc.adoc
```diff
@@ -0,0 +1,341 @@
...
+|====
+
+[[migrate-stylesheet]]
+=== Default HTML stylesheet
+The {ap} and Acidoctor stylesheets are completely compatible, because their formatting is based on the same role attributes.
+If you happen to prefer the {ap} stylesheet, you can use it by copying it from the {ap} "stylesheets" directory and telling {ad} to use it with `-a stylesheet=asciidoc.css`.
+
+// I may be doing something wrong. Can't believe no-one else has had this problem!
+
+IMPORTANT: The {ad} stylesheet (unlike the {ap} stylesheet) loads its font based icons from the web, therefore does not work well offline. To solve this, do not use inline icons <<inline-icons>>.
```
mojavelinux: I don't mind the warning about the use of web fonts, but the recommendation that follows isn't correct. There is a way to use local assets and that should be the focus of the recommendation.

mojavelinux:
Overall, an excellent start. If you can address my inline comments, I'll do a more in-depth review. This will no doubt be a very useful appendix and complete our migration story.

jaredmorgs:
Great stuff here.

SGML may still be used by archaic Defense departments. Hopefully not
though. It is horrible. Standard and Generalized should not be used
anywhere near "markup language" 😷

On Tue, 8 Mar 2016, 15:45 Dan Allen <notifications@github.com> wrote:

> Overall, an excellent start. If you can address my inline comments, I'll
> do a more in-depth review. This will no doubt be a very useful appendix and
> complete our migration story.
>
> —
> <https://github.com/asciidoctor/asciidoctor.org/pull/532#issuecomment-193618372>
> .
>
-- 

Sent from Mobile.

docs/_includes/attr-builtin.adoc
```diff
@@ -570,6 +570,13 @@ Mutually exclusive with the `notitle` attribute.
 |{y}
 |<<stem-in>>
 
+|tabsize
+|Converts tabs to spaces in verbatim content blocks (eg listing, literal).
+|4
+|1 or more
```
mojavelinux: Technically it could be 0 or more.

docs/migrating-from-asciidoc.adoc
```diff
@@ -0,0 +1,329 @@
+// Issue #480. #434, #463 refer.
+:ap: Asciidoc Python
+:ad: Asciidoctor
```
mojavelinux: Let's use "adr" for Asciidoctor, short for "asciidoc + doctor" and "adp" for "asciidoc python". That catches the syllables too.

docs/migrating-from-asciidoc.adoc
```diff
@@ -0,0 +1,329 @@
...
+=== Command line
+
+{ad} was designed from the outset to be a drop-in replacement for {ap}. 
+For most documents therefore you can just replace the {ap} call:
+
+`{ad} mydocument.adoc`
+
+with this {ad} call:
+
+`{ad} -a compat-mode mydocument.adoc`
```
mojavelinux: You'll need to write these commands out since they are not the same as the proper name. Also, use a literal paragraph instead (with a single leading space):

```
 $ asciidoc document.adoc
```

and

```
 $ asciidoctor -a compat-mode document.adoc
```

docs/migrating-from-asciidoc.adoc
```diff
@@ -0,0 +1,329 @@
...
+
+If this doesn't work for you, check out the differences below, particularly <<migrate-deprecated>> and <<migrate-extensions>>.
+
+==== internationalisation
```
mojavelinux: Can you capitalize and use "z" instead of "s"?

docs/migrating-from-asciidoc.adoc
```diff
@@ -0,0 +1,329 @@
...
+[source]
+----
+= Title
+----
+
+|====
+
+//Dan calls this "setext" style, but this term is not used in either the ad or ap manuals, so I prefer to use the ap term "two line title"
+CAUTION: [1] {ad} accepts the two-line heading style to set the document title, but using it implicitly sets `compat-mode`.
+Therefore if you want to use the new {ad} syntax, make sure to use the one-line style for the document title.
```
mojavelinux: I'm okay with that, but let's say "single-line style" instead of "one-line style".

docs/migrating-from-asciidoc.adoc
```diff
@@ -0,0 +1,329 @@
...
+|pass:[`++`] and pass:[`+++`]
+|pass:[((Sword))] and pass:[(((Sword, Broadsword, Excalibur)))]
+
+|Underlined titles
+|Underline length must match title length +/- 2 characters
+|Underline length must match title length +/- 1 character (Underlined titles are deprecated anyway)
+
+|+ifeval::[]+ 
+|Flexible
+|Compares the values of attributes only
```
mojavelinux: The way I would describe the ifeval difference is that AsciiDoc Python evaluates Python code, whereas Asciidoctor uses basic comparator logic (sandboxed).

docs/migrating-from-asciidoc.adoc
```diff
@@ -0,0 +1,329 @@
...
+
+These non backward-compatible changes are not affected by the `compat-mode` attribute:
+
+[cols="1a,1a,2a"]
+|====
+|Feature |{ap} |{ad}
+
+|Index terms
+|pass:[`++`] and pass:[`+++`]
+|pass:[((Sword))] and pass:[(((Sword, Broadsword, Excalibur)))]
```
mojavelinux: The index term syntax isn't a difference between AsciiDoc Python and Asciidoctor so much as it is about a legacy version of AsciiDoc Python (v7) and the latest version of AsciiDoc Python (v8). Therefore, I don't think we should mention it.

docs/migrating-from-asciidoc.adoc
```diff
@@ -0,0 +1,329 @@
...
+|Delimiters do not have to match in length
+|The length of start and end delimiters for a block must match
+
+|Asciidoc cell formatting (e.g `a\|`)
+|`a` or `asciidoc`
+|`a` only
+
+|Deprecated tables
+|Yes
+|No (you don't want them anyway)
```
mojavelinux: I'm hesitant to mention deprecated tables as that is again something that was dropped in AsciiDoc Python itself.

docs/migrating-from-asciidoc.adoc
```diff
@@ -0,0 +1,329 @@
...
+|`replacements2`
+|`post_replacements`
+|Renamed.
+
+// http://www.methods.co.nz/asciidoc/userguide.html#X39
+// an ugly hack so not implemented
+|`plaintext`
+|_not implemented_
+|{ap} uses this to suppress inline substitutions and retain block indents when importing large blocks of plain text.
+{ad} deliberately does not implement it; the closest {ad} equivalent is a source block.
```
mojavelinux: I think the closest equivalent to plaintext is the pass block.

docs/migrating-from-asciidoc.adoc
```diff
@@ -0,0 +1,329 @@
...
+
+// think this should be left in for completeness
+|sgml
+|_not implemented_
+|SGML is archaic and has been replaced by XML.
+
+|showcomments
+|_not implemented_
+|In {ap}, turns single line comments into DocBook "remark" elements.
+In {ad} use a pass block.
```
mojavelinux: This information got misplaced. What we should say about showcomments is that the Asciidoctor philosophy is that comments should not be passed on to the output. That's not what they are for.

mojavelinux:
We're really starting to get there.

I think that this document should be an include. Can you move it to the docs/_includes directory?

docs/migrating-from-asciidoc.adoc
jaredmorgs: I'm not sure whether I made that point clear in the ifeval section I added
recently.

On Wed, 9 Mar 2016, 19:21 Dan Allen <notifications@github.com> wrote:

> In docs/migrating-from-asciidoc.adoc
> <https://github.com/asciidoctor/asciidoctor.org/pull/532#discussion_r55490615>
> :
>
> > +
> > +[cols="1a,1a,2a"]
> > +|====
> > +|Feature |{ap} |{ad}
> > +
> > +|Index terms
> > +|pass:[`++`] and pass:[`+++`]
> > +|pass:[((Sword))] and pass:[(((Sword, Broadsword, Excalibur)))]
> > +
> > +|Underlined titles
> > +|Underline length must match title length +/- 2 characters
> > +|Underline length must match title length +/- 1 character (Underlined titles are deprecated anyway)
> > +
> > +|+ifeval::[]+
> > +|Flexible
> > +|Compares the values of attributes only
>
> The way I would describe the ifeval difference is that AsciiDoc Python
> evaluates Python code, whereas Asciidoctor uses basic comparator logic
> (sandboxed).
>
> —
> <https://github.com/asciidoctor/asciidoctor.org/pull/532/files#r55490615>.
>
-- 

Sent from Mobile.

mojavelinux: I'm pretty sure it's clear. I did some revisions after the merge.

rockyallen:
Got confused with rebase again, so please ignore this and use PR #540 instead.

mojavelinux:
For now, I'd recommend avoiding the use of rebase while we're iterating on an issue and instead use `git merge`. I understand, though, that even that can lead to merge conflicts which are tricky to resolve. But at least you can undo a merge if things get too messed up. A rebase rewrites the history, so you can't undo it.

PR #533, opened 2016-03-06
Fix patch number of Ruby version

Mogztter:
RVM doesn't like version `2.3` (version without patch number)

> Unknown ruby string (do not know how to handle): ruby-2.3.
> ruby-2.3 is not installed.
> To install do: 'rvm install ruby-2.3'


mojavelinux:
RVM :angry:

mojavelinux:
Thanks!

PR #534, opened 2016-03-06
Issue 531 relfileprefix

rockyallen:


docs/_includes/attr-builtin.adoc
```diff
@@ -532,21 +532,19 @@ Can also be set implicity from the revision line.
 |{y}
 |<<docbook>>
 
+|outfilesuffix
+|File extension to add to output files. (including leading period).
+|_empty_
+|_empty_
+|
+|<<navigating-between-source-files>>
```
mojavelinux: I like adding outfilesuffix to the built-in attributes table since it is an attribute that can be controlled. However, I don't think we should remove it from the intrinsic attribute table since it does get a default value based on converter being used. It's one of those attributes that lives in both world. Could you restore the entry in attr-data.adoc?

docs/_includes/xref-source-to-source.adoc
```diff
@@ -21,3 +21,17 @@ We could also write the link as link:README{outfilesuffix}[README].
 The links in the generated document will now point to [path]_README.adoc_ instead of the default, [path]_README.html_.
 
 CAUTION: You probably don't want to set `outfilesuffix` to `.adoc` without the `ifdef` condition as it could result in Asciidoctor overwriting input files when you run it locally (though there's some protection against this).
+
+The `relfileprefix` attribute is prefixed to relative paths, and is often used with `outfilesuffix`.
```
mojavelinux: We should clarify here that it is prefixed to relative xref paths. It is not prefixed to all relative paths (for instance, it is not applicable for relative paths defined using the link macro).

docs/_includes/xref-interdoc.adoc
```diff
@@ -51,6 +51,7 @@ If [file]_document-b.adoc_ is included in the same master as [file]_document-a.d
 
 Now you can create inter-document cross references without the headache.
 
+[[navigating-between-source-files]]
```
mojavelinux: We only need explicit IDs if it differs from the auto-generated ID, unless you think the explicit assignment is just good practice.

rockyallen:
Changes made.
Actually I do always use explicit IDs, but removed it for consistency with the rest of the doc. Implicit ids feel too much like magic for me!

mojavelinux:
Thanks!

Merged as 444a815f868410cd9935f7ded0fe7782d28a4181 with additional revisions in 39ddeb048e5f337d3fa3f20bdb29bc2719799225.

PR #535, opened 2016-03-06
Issue 480 attributes

rockyallen:
More notes - please reject if OTT.

### DIFF
--- a/docs/_includes/attr-charref.adoc
+++ b/docs/_includes/attr-charref.adoc
@@ -2,7 +2,7 @@
 = Predefined attributes for character replacements
 
 // tag::table[]
-.Predefined attributes for character replacements^[1]^
+.Predefined attributes for character replacements^[1][2]^
 [width="75%", cols="^4l,^3l,^3"]
 |===
 |Attribute name |Replacement text |Appearance
@@ -19,11 +19,11 @@
 |\&#160;
 |{nbsp}
 
-|zwsp^[2]^
+|zwsp^[3]^
 |\&#8203;
 |{zwsp}
 
-|wj^[3]^
+|wj^[4]^
 |\&#8288;
 |{wj}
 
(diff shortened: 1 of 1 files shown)


docs/_includes/attr-charref.adoc
```diff
@@ -125,7 +125,11 @@ The character references (e.g., \&#34;) are used whenever the Unicode character
 The idea is to output text that just works.
 It's up to the converter to transform this text into something the viewer understands (for example, in the case of manpage or PDF output).
 
-^[2]^ The Zero Width Space (ZWSP) is a code point in Unicode that shows where a long word can be split if necessary.
+^[2]^ Asciidoctor does not prevent you from reassigning these attributes, but it is better to treat them as read-only.
 
-^[3]^ The word joiner (WJ) is a code point in Unicode prevents a line break at its position.
+^[3]^ You can also use any of the 252 HTML 4 character entity references, for example \&euro; gives &euro;
```
mojavelinux: This statement is not entirely true. It highly depends on the output format. What is true is that Asciidoctor passes named entities through. If you're generating HTML, then any of the named HTML 4 entities will be recognized by the viewer (i.e., browser). However, this causes problems when generating non-HTML output because it relies on the entities being defined in that toolchain. It particularly break down when creating PDF.

In general, our recommendation is to avoid using named entities except for the ones defined in XML (lt, gt, amp, quot, apos) (hence why we define them here).

Therefore, I'd change the footnote to say the following:

> Asciidoctor will pass through any named entity to the output document (e.g., \&euro;). However, it is up to the interpreter (e.g., browser) to recognize these entities. Aside from the named entities defined in XML (lt, gt, amp, quot, apos), we recommend always using the numeric character reference equivalent for the best universal support.

rockyallen:
Hadn't realized that. In that case, please ignore this PR; it almost needs a warning *against* using them!

But does this clash with section 40.4:
"The replacements substitution also recognizes HTML and XML character entity references as well as decimal and hexadecimal Unicode code points and substitutes them for their corresponding decimal form Unicode code point.
For example, to render the § symbol you could write &sect;, &#x00A7;, or &#167;. When the document is processed, replacements will replace the section symbol reference, regardless of whether it is a character entity reference or a numeric character reference, with &#167;. In turn, &#167; will render as §."

mojavelinux:
Correct, 40.4 was written with HTML output in mind. We need to make our reasoning for preferring numeric character references clear. In general, I've been taught that my whole programming career. Named entities should be avoided as they are not viewed as portable (or require a big lookup table to be lugged around).

mojavelinux:
So I agree we should say that named entities are supported (or else people would riot), but discouraged.

mojavelinux:
I think we found something that works. Thanks!

Merged as 868632e05fd3ae90c44396180ae7d8ef5d728277 with revisions in c6fe80a24f8ffaf2b960b8a4cef8f7316900c2a0.

Issue #435, opened 2015-12-11
"Compared to Markdown" Content Discussion


jaredmorgs:
http://asciidoctor.org/docs/user-manual/#compared-to-markdown is currently available to create content for.

Use this issue as a tracker for content suggestions, or feel free to provide PRs referencing this ticket to keep track of contributions.

This content is contained in https://github.com/asciidoctor/asciidoctor.org/blob/master/docs/user-manual.adoc

mojavelinux:
Here's some more good raw material on the subject:

https://github.com/graphitefriction/oscon-2013-docs-workshop/blob/master/convert-markdown-to-asciidoc.adoc

mojavelinux:
When this page goes live, post a link to it on the following issue in the GitLab CE project.

https://gitlab.com/gitlab-org/gitlab-ce/issues/10778

mojavelinux:
Here are some opinion pieces on the subject we might want to link to:

* https://medium.com/@bbirdiman/markdown-considered-harmful-495ccfe24a52
* http://ericholscher.com/blog/2016/mar/15/dont-use-markdown-for-technical-docs/

mojavelinux:
The first draft is in! Let's track revisions in dedicated issues.

mojavelinux:
I posted a link to the section in the comments of the referenced GitLab issue.

nancym:
Under the table *1.5.1. A basic comparison by example*, it says:

    * Asciidoctor also supports the Markdown syntax for this language feature.

The asterisks signifying this were hard for me to notice. Here are a few suggestions:

1. Because asterisks have special meaning in a lot of Asciidoc syntax, use a different character to signify this.
2. Put this character at the beginning of the **Language Feature** text rather than the end.
3. Maybe color this character so it stands out even more.

This table is very useful and I've printed it out and use it when writing in Markdown (e.g. now in this GitHub issue) and when I'm writing AsciiDoc. It's confusing to be bilingual with these two languages!

mojavelinux:
Please open a separate issue if you'd like to propose improvements to this section.

Issue #530, opened 2016-03-06
Create a page of leading articles about AsciiDoc and Asciidoctor


mojavelinux:
Create a page listing leading articles about AsciiDoc and Asciidoctor. The articles in this list would be things like case studies, stories, experiences, recommendations or advocacy.

The goal isn't to capture every article on the internet about AsciiDoc and Asciidoctor, but to make it a curated list of more recent and well-written entries. In other words, it would be a curated list both for us to use as reference or defense as well as for newcomers to learn from.

Proposed path: `/resources/articles/`

mojavelinux:
Here are a few articles to consider:

* [Living the Future of Technical Writing](https://medium.com/@chacon/living-the-future-of-technical-writing-2f368bd0a272)
* [Eloquent JavaScript's Build System](http://marijnhaverbeke.nl/blog/eloquent-javascipts-build-system.html)
* [AsciiDoc & AsciiDoctor first step and first thoughts](http://www.mimiz.fr/en/asciidoc-asciidoctor-first-step-and-first-thoughts) (this one might be leaning more towards a tutorial

PR #541, opened 2016-03-10
fix #34 and #435

### LINKED ISSUE #34: Write differences between AsciiDoc and Markdown documentation
How do the two syntaxes differ?
Partners with the Asciidoctor and AsciiDoc differences
May go under The Basics category

andya9:
A first draft (mostly using Dan’s own writings on a github issue) so that the section isn’t fully empty.
Please comment =D

### DIFF
--- a/docs/_includes/asciidoc-vs-markdown.adoc
+++ b/docs/_includes/asciidoc-vs-markdown.adoc
@@ -1,0 +1,259 @@
+////
+A comparison between asciidoc and markdown.
+This file is included in the user-manual document
+////
+
+The most compelling reason in choosing a lightweight markup is minimizing the technical concepts that content authors must grasp in order for them to be immediately productive.
+
+A popular choice is Markdown. The main advantage of Markdown lies in its small syntax: its manual could be its own cheatsheet. But this same advantage is also its greatest fault: as soon as authors need something slightly more complex (a table, footnotes, including YouTube videos, ecc) they find themselves manually typing in raw html, or looking for extensions to the language (so that Markdown has now become a mess of different implementations, called "`flavours`", that prevent it to be truly universal).
+
+AsciiDoc is an excellent alternative in that it offers a syntax more concise than (or as concise as) Markdown, while being by far more flexible and more powerful without requiring the use of HTML or "`flavors`" to get essential syntax like tables, definition lists, admonitions (tips, notes, warnings, etc) and table of contents.
+
+==== A basic comparison by example
+
+[cols=3]
+|===
+| Formatting
+| Markdown
+| AsciiDoc
+
+| Bold (word)
+a|
+[source, markdown]
+----
+**bold**
+----
+a|
+[source, asciidoc]
+----
+*bold*
+----
+
+| Bold (first letter)
+a|
+[source, markdown]
+----
+**b**old
+----
+a|
+[source, asciidoc]
+----
+**b**old
+----
+
+| Italic (word)
+a|
+[source, markdown]
+----
+*italic*
+----
+a|
+[source, asciidoc]
+----
+_italic_
+----
+
+| Italic (first letter)
+|n/a
+a|
+[source, asciidoc]
+----
+__i__talic
+----
+
+| Literal (word)
+a|
+[source, markdown]
+----
+`literal`
+----
+a|
+[source, asciidoc]
+----
++literal+ or `literal`
+----
+
+| Literal (first letter)
+a|
+[source, markdown]
+----
+`l`iteral
+----
+a|
+[source, asciidoc]
+----
+++l++iteral
+----
+
+| Link with text
+a|
+[source, markdown]
+----
+[AsciiDoc](http://asciidoc.org)
+----
+a|
+[source, asciidoc]
+----
+http://asciidoc.org[AsciiDoc]
+----
+
+| Image w/ Alt Text
+a|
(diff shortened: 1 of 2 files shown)


docs/_includes/asciidoc-vs-markdown.adoc
```diff
@@ -0,0 +1,259 @@
...
+| Literal (word)
+a|
+[source, markdown]
+----
+`literal`
+----
+a|
+[source, asciidoc]
+----
++literal+ or `literal`
```
mojavelinux: Let's use the modern syntax only, so drop the `+literal+` one.

docs/_includes/asciidoc-vs-markdown.adoc
```diff
@@ -0,0 +1,259 @@
...
+| Literal (first letter)
+a|
+[source, markdown]
+----
+`l`iteral
+----
+a|
+[source, asciidoc]
+----
+++l++iteral
```
mojavelinux: Use double backtick here instead of double plus.

docs/_includes/asciidoc-vs-markdown.adoc
```diff
@@ -0,0 +1,259 @@
...
+| Image w/ Alt Text
+a|
+[source, markdown]
+----
+![Logo](/images/logo.png)
+----
+a|
+[source, asciidoc]
+----
+image:logo.png[Logo]
```
mojavelinux: Make it clear that this is an inline image. We may want another entry for "Block Image w/ Alt Text" and say that there is no equivalent in Markdown.

docs/_includes/asciidoc-vs-markdown.adoc
```diff
@@ -0,0 +1,259 @@
...
+You can see that AsciiDoc has the following advantages over Markdown:
+
+* it uses the same number of markup characters or less when compared to Markdown in nearly all cases
+* it employs a consistent formatting scheme
+* it can handle all permutations of nested inline (and block) formatting, whereas Markdown often falls down
+* it handles cases that Markdown doesn't (like a proper approach to inner-word markup and source code blocks)
++
+For example:
++
+*Definition Lists in AsciiDoc*
```
mojavelinux: Let's use subsections for "Definition Lists" and "Tables".

docs/_includes/asciidoc-vs-markdown.adoc
```diff
@@ -0,0 +1,259 @@
...
+
+| 1
+| 2
+| 3
+|===
++
+Markdown Extra supports tables and definition lists; but it’s not Markdown.
+Also, unlike Markdown Extra, Asciidoc can apply formatting to cells.
+
+Asciidoctor, which is used on GitHub, also emulates "`the good parts`" of the Markdown syntax, like headings, blockquotes and fenced code blocks: migration is fairly easy.
```
mojavelinux: Let's add a reference to the Markdown support. For now, let's link to http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#markdown-compatibility

mojavelinux:
A fantastic start! I added some inline comments.

The next question to ask is, what's missing? I think it's fine to merge before it's perfect, but it's an important step for identifying the next action.

mojavelinux:
I'd like to include a statement for the intro similar to the one proposed in #34.

> I think it is important to remember that AsciiDoc was initially designed to be a plain text subset of DocBook metadata, and I think that is still a primary use case for AsciiDoc. However, it should also be simple enough to be a "better" version of markdown (and I would argue it already is), thereby fulfilling the the re-usability of the written text. In that regard, there maybe a good argument for a minidoc that focuses on showing markdown users how to do the same "formatting" in AsciiDoc, but it should emphasise that AsciiDoc has much more to offer.

It's important to understand that AsciiDoc was born out of DocBook. That gives people confidence that it isn't just shooting in the dark rather than just trying to patch needs like Markdown.

We could use less words, though.

mojavelinux:
I think there are a few missing comparisons that are covered in the following page:

https://github.com/graphitefriction/oscon-2013-docs-workshop/blob/master/convert-markdown-to-asciidoc.adoc

I also think we should link back to issue #34 for people who want to dive deeper. Even though the issue will be closed by this PR, it still holds some detail that isn't appropriate for adding to the user manual.

andya9:
I implemented some fixes following your comments. As for what is missing, maybe some notes on tool availability?

mojavelinux:
Thanks!

I think the discussion of tooling is out of scope. The focus of this section is really on the syntax, IMO.

andya9:
You are right :smile:

andya9:
I added two more asciidoc only features: split source and custom css classes

andya9:
@mojavelinux What do you think of [this link by Scott Chacon](https://medium.com/@chacon/living-the-future-of-technical-writing-2f368bd0a272#.wv00yvxzf)? I believe it’s a beautiful proof of asciidoc’s value from a well-known actual technical writer; IMO, we should link to it; where?

mojavelinux:
See #530.

andya9:
Aah, I missed it :smile:

mojavelinux:
Thanks!!

Merged as 116a0ee5e3e3735da9fe39c815275b4d6d83cc1e with additional revisions in b95aa272f05fc8ceda2f656803b6149b66cb8d9f.

PR #546, opened 2016-03-13
fix #544

### LINKED ISSUE #544: Document how to activate the AsciiMath interpreter when converting to DocBook
Asciidoctor supports converting AsciiMath to MathML in the DocBook converter. This requires at least Asciidoctor 1.5.4 and the asciimath gem. Document this feature as a subsection of http://asciidoctor.org/docs/user-manual/#activating-stem-support in the user manual.

For detailed documentation, see https://github.com/asciidoctor/asciidoctor-fopub/issues/49#issuecomment-166061234.

andya9:
I didn’t understand the jeuclide reference, so I didn’t deal with it

### DIFF
--- a/docs/_includes/stem.adoc
+++ b/docs/_includes/stem.adoc
@@ -36,6 +36,11 @@ If you want to use the LaTeX interpreter by default, assign `latexmath` to the s
 
 The html backend will just use mathjax client-side to render math notation, whereas the docbook backend will use it to convert math to mathml.
 
+[NOTE]
+====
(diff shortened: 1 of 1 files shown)


mojavelinux:
Thanks!

By default, FOP does not process MathML. The JEuclid plugin for FOP is what interprets the MathML that the asciimath gem puts into the DocBook.

I think all we need to say about this is:

"fopub is configured to use the JEuclid FOP plugin. When fopub generates the PDF, this plugin interprets the MathML that the asciimath gem adds to the DocBook file. As a result, stem expressions in the AsciiDoc file that are written in AsciiMath will render as expected in the generated PDF."

Can you add that, perhaps as a NOTE admonition?

mojavelinux:
Thanks @andya9!

Merged as e9d44766f0e19d773a5cd2b788fb63e53e2c4c19 with subsequent revisions in 7e5d00187c2eb273f9e7b732b4f55da9d8b8d992.

andya9:
Beautiful!

PR #547, opened 2016-03-13
fix #516

### LINKED ISSUE #516: Migrate all files in the docs folder to the modern AsciiDoc syntax
All the files in the docs folder (in particular, the user manual) need to be upgraded to the modern AsciiDoc syntax supported by Asciidoctor. In other words, we want to eliminate the use of compatibility mode (`compat-mode`). Right now, the user manual uses mixed modes, making it really hard for contributors to know which syntax to use.

andya9:
Just a small start. I wonder: if we convert to docbook, and then back again with the new syntax, maybe it’s faster then doing it manually? (is it even possible?)


mojavelinux:
This is going to have to be done all at once because it has to coincide with a change to the compat-mode attribute. The alternative is just too confusing, which is to try to toggle the compat-mode attribute back and forth around the sections that have been converted.

>  if we convert to docbook, and then back again with the new syntax, maybe it’s faster then doing it manually? (is it even possible?)

Not at the moment. We still aren't at the point where we have perfect round tripping.

andya9:
Ok, should I close this?

mojavelinux:
I'd say just keep going. I just can't easily merge it until all the files are converted.

There is one alternative approach. Instead of working through the include files alphabetically, you can work through the includes in the user-manual.adoc in sequence. Then, you should be able to see how you can keep moving the `compat-mode` setting down as you make progress.

mojavelinux:
The focus of this issue is only on the files that the user-manual includes. We can then deal with all the other files later. It's the mixed mode in the user manual that is throwing a wrench in collaboration.

andya9:
Ok. My problem is that I’m new to asciidoc, so I don’t know all of its syntax: for example in attr-miss.adoc, I don’t understand this; to me, it seems sort of like a nested literal, but it wouldn’t make sense:
```
  `+{bogus}+`
```
So I fear that I’ll change something that I shouldn’t..

mojavelinux:
This is how you make monospaced text that doesn't get interpolated or formatted. In other words, it's escaped monospace.

Have you seen http://asciidoctor.org/docs/migration/#migration-scenarios?

Don't worry about making mistakes. I'll be sure to QA everything since it needs to all be tested anyway.

andya9:
Ah, thank’you! Are those 4 user cases the only ones to address?

mojavelinux:
Yep!

andya9:
It *should* be fine now!

docs/_includes/factory-setup.adoc
```diff
@@ -39,11 +39,11 @@ Once you have the gems installed, you can build the stylesheets.
...
-. Run Compass's +compile+ command.
+. Navigate to the Asciidoctor stylesheet factory directory on your system.
+. Run Compass's `compile` command.
 
  $ compass compile
 
-The stylesheets are compiled from the Sass source files in the [path]'sass/' folder and written to the [path]'stylesheets/' folder.
-You can reference the stylesheets in [path]'stylesheets/' from your HTML file.
+The stylesheets are compiled from the Sass source files in the [path]__sass/__ folder and written to the [path]__stylesheets/__ folder.
+You can reference the stylesheets in [path]__stylesheets/__ from your HTML file.
```
mojavelinux: You only need single underscores around the terms here to make them italic.

```
in the [path]_sass/_ folder ... to the [path]_stylesheets/_ folder.
```

docs/_includes/static-awe.adoc
```diff
@@ -5,15 +5,15 @@ Included in:
 ////
 
 Awestruct defines a set of default attributes that it passes to the API in its [path]'/default-site.yml' file.
-One of the attributes in that configuration is +imagesdir+. 
-The value there is set to [path]'/images'. 
+One of the attributes in that configuration is `imagesdir`.
+The value there is set to [path]'/images'.
```
mojavelinux: This legacy italic got missed.

```
The value there is set to [path]_/images_.
```

docs/_includes/static-awe.adoc
```diff
@@ -5,15 +5,15 @@ Included in:
...
 
-Fortunately, there is one additional place you can override the attribute. 
-This gives you the opportunity to set your own default and to flip the precedence order so that the document wins out. 
-If an attribute value that is passed to the API ends with an +@+ symbol, it makes that assignment have a lower precedence than an assignment in the document.
+Fortunately, there is one additional place you can override the attribute.
+This gives you the opportunity to set your own default and to flip the precedence order so that the document wins out.
+If an attribute value that is passed to the API ends with an `@` symbol, it makes that assignment have a lower precedence than an assignment in the document.
 
-You can define attributes you want to pass to the API in the [path]'_config/site.yml' file. 
+You can define attributes you want to pass to the API in the [path]'_config/site.yml' file.
```
mojavelinux: This legacy italic got overlooked.

```
in the [path]__config/site.yml_ file.
```

docs/_includes/text-mono.adoc
```diff
@@ -112,7 +112,7 @@ While the previous examples will work in Asciidoctor 1.5.0, the last two are now
 To use the transitional syntax, add the role "x-" in front of the legacy syntax to indicate that you want Asciidoctor 1.5.0 to use the old behavior. Of course, Asciidoctor 0.1.4 already understands the old syntax.
 
 * [x-]\`Use {asciidoctor-version} to print the version of Asciidoctor` (attribute is _not_ replaced)
-* [x-]+The version of Asciidoctor is {asciidoctor-version}+ (attribute _is_ replaced)
+* [x-]`The version of Asciidoctor is {asciidoctor-version}` (attribute _is_ replaced)
```
mojavelinux: This one we don't want to replace because it is showing how to use the backwards-compatible syntax.

docs/_includes/text-mono.adoc
```diff
@@ -97,9 +97,9 @@ There are three scenarios that are affected by this syntax change.
 
 Let's first consider the legacy syntax used in each of these examples.
 
-* +monospaced normal text+ or \`monospaced normal text`
+* `monospaced normal text` or \`monospaced normal text`
 * \`Use {asciidoctor-version} to print the version of Asciidoctor` (attribute is _not_ replaced)
-* +The version of Asciidoctor is {asciidoctor-version}+ (attribute _is_ replaced)
+* `The version of Asciidoctor is {asciidoctor-version}` (attribute _is_ replaced)
```
mojavelinux: We don't want to replace these because they are showing how to use the legacy syntax.

mojavelinux:
Great work! I added a few inline comments to address. I'm really looking forward to merge this!

mojavelinux:
The text-mono section is tricky since it is discussing both the new and legacy syntax. I'll need to go through that one carefully because it's confusing even for me :)

andya9:
Done! (Sorry for the mess with commits, I missed a couple of your comments)

mojavelinux:
Thanks! Great job!

Don't worry about the commits. I tend to rebase locally for larger changes like this one when I review, so it ends up nice and tidy in the end.

mojavelinux:
If you search for `path]'`, you'll find a handful more of instances where single quotes are being used for italics that need conversion. Those are tough to find because single quotes are so common, but hopefully that search string narrows it down.

andya9:
Done; are there other patterns with ' similar to this to fix?

mojavelinux:
I sent a PR to your PR with a few additional changes and removed the compat-mode attribute from the master file. As far as I can tell, we've caught all the cases.

andya9:
Perfect! :sparkles:

mojavelinux:
Merged as 295c1cacf4fe1d18aa67c8dd11679ee0294a19e2. Thanks!

Issue #559, opened 2016-03-22
Fix broken links in user manual


rockyallen:
Caused by removal of old appendix A and new migration from asciidoc python section

PR #560, opened 2016-03-22
Resolves issue #559 broken links

rockyallen:


mojavelinux:
Excellent! However, I don't think we should use a custom label on the xref unless it is different from the target section title. Could you simplify to `<<<<asciidoctor-vs-asciidoc>>`. Though, thinking about it more, maybe we should also fix the ID so it is more appropriate. Perhaps `<<migrating-from-asciidoc-python>>`?

rockyallen:
Done

mojavelinux:
Nice! Merged as da5686ebd942076cf7dc8e2ac581d1ff0d629c59.

You'll notice I dropped the word "issue" from the git commit message. That seems to confuse the GitHub commit parser so it doesn't automatically close the issue. In order for that to happen, the wording must be `resolves #xyz ...`.

PR #563, opened 2016-03-26
Resolves #561, #560 counters

### LINKED ISSUE #561: Explain counter attributes better
Currently only an example with no explanation. These notes are based on testing with 1.5.4 but are not all required.

Any document attribute can be used as a counter, but it is usual to use a special syntax to define them:

{counter:item_number:9} 

sets a document attribute item_number and assigns it value 9, then references it, ie it is shorthand for:

:item_number: 9
{item_number}

The initial value of the counter should[1] be a decimal integer (1 or more digits), single lower case letter, or single upper case letter.

If the initial value is not specified, the initial value is 1. Ie if item_number is not already set, {counter:item_number} is shorthand for

:item_number: 1
{item_number}

But if item_number is already set, the same expression {counter:item_number} increments the attribute value then references it.

"increment" means add 1 if the current value is an integer, or choose the next letter of the (ASCII) alphabet if it is anything else.
Therefore the number sequence runs 1,2,3....999....
but the letter sequence runs a,b,c,...x,y,z,{,|... or A,B,C,...,X,Y,Z,[,\... so isn't useful for more than 26 items.

{counter2:item_number} increments the attribute value, but does not reference it.

[1] It is not an error to have an initial value that does not match this, but the result will not be useful: for example the sequence starting "frogs" is "frogs","g","h"...

rockyallen:


docs/_includes/counter.adoc
```diff
@@ -1,9 +1,66 @@
...
+Which gives:
+
+{counter:i}. Apples, {counter:i}. Oranges, {counter:i}. Pears.
+
+The counter is incremented and displayed every time `counter:` is used.
+To skip an item in the sequence (i.e. increment the counter without displaying it), use `counter2` instead:
+
+ {counter2:item_number}
+
+To display the current value of the counter without incrementing it, reference it:
```
mojavelinux: I think this thought needs to be completed. Perhaps:

"To display the current value of the counter, but not increment it, refer to it using the normal attribute reference syntax:"

docs/_includes/counter.adoc
```diff
@@ -1,9 +1,66 @@
...
+
+ {counter:para_number} This paragraph is number {para_number}.
+
+NOTE: "increment" means add 1 if the current value is an integer, or choose the next character of the latin alphabet if it is anything else.
```
mojavelinux: I'd prefer this as:

"increment" means advance to the next value in the sequence.
If the counter value is an integer, add 1.
If the counter value is a character, move to the next character in the Latin alphabet.

docs/_includes/counter.adoc
```diff
@@ -1,9 +1,66 @@
...
+- user-manual: Counters
 ////
 
+// See issue #550, #561, https://github.com/asciidoctor/asciidoctor/issues/1559
+// Focus on simple use case.
+// {counter:name:start_value} is only for first use.
+// document attributes and counters are NOT the same thing, but modifying a document attribute with the same name as the counter modifies the counter at the same time.
+
+Counters are used to generate sequences of numbers or latin characters.
+Possible uses are to create inline itemised lists, paragraph numbers, or serial numbers.
```
mojavelinux: Can you write "itemised" as "itemized"? That aside, great example!

docs/_includes/counter.adoc
```diff
@@ -1,9 +1,66 @@
...
+
+Counters are created and used by prefixing the counter name with `counter:` and surrounding it with curly brackets.
```
mojavelinux: I feel like a parallel to an attribute references is missing here. I'd like it to say:

Counters are a special type of document attribute.
Counters are declared and displayed using a modified attribute reference in which the attribute name is prefixed with `counter:`.
Since they are attributes, counter names follow the same rules as attribute names.

(Move discussion of the start value to the next paragraph).

mojavelinux:
Thanks @rockyallen!

Merged as 531f83882a280bf1db7b762e74e9706fd16c19d5 with additional revision in 5b5d38860dfcf17d01b01149d375673fcf959bd2.

PR #570, opened 2016-04-13
Updated documentation for AsciiDoctorJ

ygra:
Taken from AsciiDoctorJ's readme, which was two versions ahead already.

Fixes asciidoctor/asciidoctorj#445.


mojavelinux:
Thanks for doing this @ygra!

PR #576, opened 2016-04-18
resolves #562 API and CLI options

### LINKED ISSUE #562: Create a catalog of options
As a parallel to the catalog of (document) attributes, the user manual should also have a catalog of options. Options are used to control how the processor itself is configured (as opposed to the content). Options handle the following tasks:

* Specify the converter to use
* Set the safe mode
* Require additional libraries
* Control the path from which libraries and other resources are loaded
* Tune the verbosity of messages
* Set the input and output files or directories
* Set the base directory

Most options are available when using both the CLI and the API, though each interface also has unique options.

rockyallen:
* Cant find any documentation about the API options marked ?
* Had to copy and paste the CLI, don't know how to include from foreign project


mojavelinux:
This is a very nice start!

I think we should mark the API option names with a leading colon (e.g., `:attributes`). That makes it clear that it is a Ruby symbol and not a string. This is consistent with the way that keys are documented in the Ruby API. See http://ruby-doc.org/core-2.3.0/IO.html#method-c-new-label-Options

(Astute Ruby developers will know that you can define a symbol key by putting the colon at the end of the key name, but I don't think we need to point that out).

docs/_includes/api-options.adoc
```diff
@@ -0,0 +1,149 @@
...
+|`-b`, `--backend`
+
+|base_dir
+|Base directory containing the document and resources.
+|The directory containing the source file, or the working directory if the source is read from a stream.
+|_path_
+|`-B`, `--base-dir`
+
+|catalog_assets
+|?
```
mojavelinux: catalog_assets tells the parser to store references to all assets (images and links) in the reference table available via the references property on the document AST object. (Normally only ids, footnotes and indexterms are captured). This is still a primitive and experimental feature but may be useful for early adopters to address specific use cases.

docs/_includes/api-options.adoc
```diff
@@ -0,0 +1,149 @@
...
+|
+|
+|_none_
+
+|compact
```
mojavelinux: Please remove compact. We aren't supporting this option any longer.

docs/_includes/api-options.adoc
```diff
@@ -0,0 +1,149 @@
...
+|Not used
+|false
+|true or false
+|_none_
+
+|converter
+|?
```
mojavelinux: The converter option accepts an instance of a Converter and is used in place of the converter that is automatically resolved from the backend value.

The default value is nil (not set). The allowed value is a Converter instance.

docs/_includes/api-options.adoc
```diff
@@ -0,0 +1,149 @@
...
+|`-d`, `--doctype`
+
+|eruby
+|Specifies the eRuby implementation to use for executing the custom converter templates written in ERB.
+|erb
+|erb, erubis
+|`-e`, `--eruby`
+
+|extensions
+|List of file extensions for asciidoc source files.
```
mojavelinux: Nope. The extensions options is a Ruby block that registers (and possibly defines) Asciidoctor extensions for this instance of the processor.

The default value is nil. The allowed value is a Ruby block that conforms to the Asciidoctor extensions API.

docs/_includes/api-options.adoc
```diff
@@ -0,0 +1,149 @@
...
+|ad, adoc, asciidoc 
+|
+|_none_
+
+|extensions_registry
+|?
```
mojavelinux: The extensions_registry option is one level up from the extensions option. Instead of providing a Ruby block containing extensions to register, this option lets you replace the extension registry itself, giving you more control over how extensions are registered for this processor.

The default value is nil. The allowed value is an Asciidoctor::Extensions::Registry instance.

(Obviously, this is a pretty advanced use case. Details will be covered in the planned guide for extension writers).

docs/_includes/api-options.adoc
```diff
@@ -0,0 +1,149 @@
...
+|`-s`, `--no-header-footer`
+
+|mkdirs
+|If true, allows {adr} to create output directory if necessary.
+|false
+|true, false
+|_none_
+
+|parse
+|?
```
mojavelinux: This controls whether the document is parsed when loaded (when passed to the load or load_file API). By default, the document is parsed eagerly. This option lets you defer parsing until the parse method is explicitly invoked.

The default value is true (technically nil, but nil is treated as true). The allowed values are true and false.

docs/_includes/api-options.adoc
```diff
@@ -0,0 +1,149 @@
...
+|_none_
+
+|safe
+|Safe mode.
+|SECURE
+|See CLI
+|`-S`, `--safe-mode`
+
+|source_map
+|?
```
mojavelinux: The sourcemap option (not source_map) keeps track of the file path and line numbers for each block that is parsed. This is useful for tooling applications where the association between the converted output and the source file is important.

docs/_includes/api-options.adoc
```diff
@@ -0,0 +1,149 @@
...
+|
+|
+|_none_
+
+|template_cache
+|?
```
mojavelinux: The template_cache option enables or disables the built-in cache used by the template converter when reading the source of template files. This is only relevant if the template_dirs option is specified.

The default is true. The allowed values are true and false.

docs/_includes/api-options.adoc
```diff
@@ -0,0 +1,149 @@
...
+|
+|
+|_none_
+
+|template_dir
```
mojavelinux: The template_dir option is deprecated in favor of template_dirs, which accepts an array of paths.

docs/_includes/api-options.adoc
```diff
@@ -0,0 +1,149 @@
...
+The gem with the same name as the engine will be loaded automatically.
+This name is also used to build the full path to the custom converter templates.
+|_none_
+
+Auto-detected based on the file extension of the custom converter templates found.
+|_name_
+|`-E`, `--template-engine`
+
+|template_engine_options
+|?
```
mojavelinux: The template_engine_options are low-level options passed to the template engine. This option borderlines on being internal and is recommended only for advanced users.

(You can see an example in the Bespoke.js converter at https://github.com/asciidoctor/asciidoctor-bespoke/blob/v1.0.0.alpha.1/lib/asciidoctor-bespoke/converter.rb#L24-L28).

docs/_includes/ruby-cli-options.adoc
```diff
@@ -0,0 +1,130 @@
...
+  Document type: _article_, _book_, _manpage_ or _inline_.
+  Sets the root element when using the _docbook_ backend and the style class on the HTML body element when using the _html_ backend.
+  The _book_ document type allows multiple level-0 section titles in a single document.
+  The _manpage_ document type enables parsing of metadata necessary to produce a manpage.
+  The _inline_ document type allows the content of a single paragraph to be formatted and returned without wrapping it in a containing element.
+  Defaults to _article_.
+
+=== Rendering Control
+
+*-C, --compact*::
```
mojavelinux: Please drop mention of the compact flag.

mojavelinux:
I provided descriptions where missing. If you incorporate those, then we'll be ready to merge the first version.

mojavelinux:
Nice work!

Merged as b4601d20bc456d36067bfd4da4052d8a175d54c2 with some additional revisions.

PR #579, opened 2016-04-22
Resolves #568 character replacements

### LINKED ISSUE #568: Clarify when replacements are used and how to escape them
ref https://github.com/asciidoctor/asciidoctor/issues/1717

Add note to 40.4: "In most cases, replacements can be escaped by adding a leading \ character (e.g., \--)" [although I tested all of them and didn't find any exceptions]

Add note in table cell for --: "Only replaced if it is a word, ie surrounded by white space, line start, or line end." [Why? Seems an arbitrary decision - feels more like a bug]

rockyallen:


mojavelinux:
Super! Thanks @rockyallen!

PR #583, opened 2016-05-11
Resolves #542 import from word (attempt 2)

### LINKED ISSUE #542: "Convert MS Word to AsciiDoc" content discussion
Use this issue as a tracker for content suggestions, or feel free to provide PRs referencing this ticket to keep track of contributions.

This content is contained in https://github.com/asciidoctor/asciidoctor.org/blob/master/docs/user-manual.adoc.

rockyallen:


AlexanderZobkov:
@rockyallen , thank you for addressing the comments. The section looks good to me now and I would recommended to merge the PR.

Regarding Ant fragments in the section, I would suggest to replace or accompany them with correponding unix shell one-lines. For example, the following Ant fragment: 
```
<replaceregexp
    file="${outfile}"
    match="\[\[_.*]]"
    replace=""
    flags="g"/>
```
can be done with the following one-liner:
```
perl -W -pe  's!\[\[_.*]]!!g' -i file.adoc
```
IMHO, it's more easier to reuse (and adapt when needed) shell one-liners comparing to Ant tasks. 
For Ant tasks you need to: download and install Ant, create a build file, populate it with tasks and deal with XML (escape <<>>). Unix shell is more accessible (even on Windows, by using Cygwin or by using experimental "Windows Subsystem for Linux" on Windows 10.)

And a small typo in docs/user-manual.adoc: +== Convert MicroSoft Word to AsciiDoc 
It probably should be "Microsoft"

AlexanderZobkov:
@mojavelinux, What do you think about the proposed section?

docs/_includes/migrate-from-msword.adoc
```diff
@@ -0,0 +1,234 @@
+// == Importing from MicroSoft Word
+:experimental:
+:url-pandoc: http://pandoc.org
+:url-ant: http://ant.apache.org/
+:url-google-asciidoc: https://chrome.google.com/webstore/detail/asciidoc-processor/eghlmnhjljbjodpeehjjcgfcjegcfbhk/
+:url-google-asciidoc-source:  https://github.com/Mogztter/asciidoc-googledocs-addon/
```
mojavelinux: Could you use the prefix "uri-" instead of "url-"?

docs/_includes/migrate-from-msword.adoc
```diff
@@ -0,0 +1,234 @@
...
+|Figure and table captions
+|Normal paragraph
+
+|Any other MS Word styled paragraphs
+|Normal paragraph
+
+|Embedded images
+|Yes
+
+|Character formating (bold, underline and italic)
```
mojavelinux: "formating" should be "formatting"

docs/_includes/migrate-from-msword.adoc
```diff
@@ -0,0 +1,234 @@
...
+
+|Linked (not embedded) images
+|Ignored
+
+|Vector graphics (MS Word "insert shape")
+|Ignored
+
+|====
+
+=== Optimised Pandoc
```
mojavelinux: Could you use the spelling "Optimized"?

docs/_includes/migrate-from-msword.adoc
```diff
@@ -0,0 +1,234 @@
...
+    match="Figure (\d?)"
+    replace="&lt;&lt;fig-\1&gt;&gt;"
+    flags="g"/>
+----
+
+=== Google Docs
+Google Docs can already upload and edit MS Word docx files.
+With this addon from Guillaume Grossetie: {url-google-asciidoc}[AsciiDoc Processor]
+you can copy and paste part or all of the document from Google Docs as AsciiDoc text. 
+The features that it can handle seem to be slightly fewer than Pandoc, but expect further development.
```
mojavelinux: We should probably say "substantially fewer" and that this tool is still in the early stages. We don't want to provide false hope and have people disappointed. In time, we can strengthen the message.

docs/_includes/migrate-from-msword.adoc
```diff
@@ -0,0 +1,234 @@
...
+
+|====
+
+=== Optimised Pandoc
+
+The basic usage above is fine for one-off imports.
+If you have lots to do, it is worth while cleaning the input document first, and automating the post-conversion tidy up.
+
+// Is this egg sucking? I have just told the reader what is and isn't implemented, if he can drive Word he doesn't need me to tell him what to do.
+// Are the comments helpful? Should they go into the text?
```
mojavelinux: I actually kind of like this section. It gives the reader the confidence and know how in making a good export. I expect that over time others will discover additional tricks that we can add here. We may also pare down the list as we find things that weren't helpful after all.

docs/user-manual.adoc
```diff
@@ -1766,6 +1766,10 @@ TIP: Use {uri-ad-org-issues}/462[Issue 462] to drive development of this section
 
 include::{includedir}/convert-from-confluence-xhtml.adoc[]
 
+== Convert MicroSoft Word to AsciiDoc
```
mojavelinux: Can you lowercase the "s" in Microsoft?

mojavelinux:
I agree with @rockyallen that Ant is probably the most portable solution. While Perl can certainly handle these tasks, it's easier for people to run Ant cross platform.

Ideally, though, I think these scripts should be moved into a repository so that it isn't necessary to copy and paste them out of the documentation. I don't think it's a good idea to maintain code as part of the documentation. Instead, the documentation should include fragments from the code.

@rockyallen would you be willing to create a project on GitHub to store these scripts? Perhaps we could just call it asciidoc-conversion-tools (or helpers).

rockyallen:
OK. Move scripts to new project, Americani(s/z)e spellings. I am a bit vague on Github repositories - if I make one in my Github area, will you be able to move ti to where you want when it is ready?

AlexanderZobkov:
@mojavelinux , Ant probably allows to run things in cross platform manner. However in case of asciidoctor ecosystem, probably another tool needs to chosen. The command line tools that are provided in Asciidoctor repository are mostly based on ruby and/or unix shell. For example: docbookrx, asciidoctor-pdf, asciidoctor-epub3, asciidoctor-fopub. Introducing ant based tool, we are forcing people to learn, install and configure ant + java just to invoke pandoc + perfom several replacments. I would vote for a tool that is based on unix shell (preferable) or ruby (less preferable). Ant is a good tool but requires more manual steps to run.and adds burden, in this particular case.
 
I have no intention to argue, just convince both of you that something is based on unix shell or ruby is probably a better solution in this particular case. Any decision/opinion is ok for me :smile:

jirutka:
I agree with @AlexanderZobkov except 1) I'm convinced that Ant is _not_ a good to at all (it's basically a scripting in XML, a nightmare), 2) I'd prefer Ruby to shell, because it works even in Windows without need to install cigwin.

mojavelinux:
If we make a dedicated project, then I'm more inclined to use a tool that aligns with the ecysostem. In this case, I agree with @AlexanderZobkov and @jirutka that it should be a Ruby script.

@rockyallen If you commit the Ant build to a repository, then I'm sure one of us will be happy to convert it to Ruby for you. And yes, if you create a repository in your own personal space on GitHub, we can still move it to the Asciidoctor organization.

What I'm still not sure about is whether we want a separate repository for this script or whether it should live in an existing repository. One that comes to mind is the extensions-lab. But before we decide on that, I'd say just commit it to a personal repository and we can at least reference that until we decide.

mojavelinux:
Should we merge the PR then update the file once we migrate the scripts to a new location, or shall we wait?

AlexanderZobkov:
My suggestion is the following:
1. Address wording comments provided by @mojavelinux
2. Replace Ant task snippets with perl analogous (Sorry, I like unix shell stuff :smile: )
3. Review and merge the section 
4. Submit a new issue: add a ruby script to asciidoctor-extensions-lab/scripts/ to automate/ease migration from MS Word to asciidoctor, update  the section with reference to the script to be added.

wdyt?

AlexanderZobkov:
@rockyallen , are you willing to address the wording comments? I would like take points 1 and 2 from the plan.

AlexanderZobkov:
@mojavelinux and @rockyallen, Review comments are addressed in https://github.com/asciidoctor/asciidoctor.org/pull/601.

mojavelinux:
I merged #601. If we need to make additional changes, let's handle them using follow-up issues and/or pull requests.

Great job everyone! We're off to a good start!

PR #584, opened 2016-05-13
Add 'chapter-label' in the user-manual

jmini:
http://discuss.asciidoctor.org/How-can-we-change-quot-chapter-quot-td4667.html

### DIFF
--- a/docs/_includes/attrs-builtin.adoc
+++ b/docs/_includes/attrs-builtin.adoc
@@ -92,6 +92,13 @@ Useful if you want to store the output in a source code control system as it pre
 |
 |<<admonition>>, <<customizing-labels>>
 
+|chapter-label
+|Sets the prefix set bevor Level 1 sections title, when the doctype is book (pdf backend).
+|Chapter
+|_any_
+|
+|<<book-parts>>, <<sections>>, <<customizing-labels>>
+
 |example-caption
 |Sets the text used to label example blocks.
 |Example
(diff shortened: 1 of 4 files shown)


mojavelinux:
I do want to add this, but I want to hold off until Asciidoctor PDF is out of alpha. The other reason is that you'll notice that the attribute name is `chapter-label` and not `chapter-caption`. We need to change all the `-caption` attributes to be `-label` because "caption" is the wrong term here. For instance, `example-label`. But I still haven't figured out how to do this in a way that doesn't break stuff.

jmini:
@mojavelinux I really respect your opinion and your vision, but I think you wrong on this one.

I know that you plan to [rename `*-caption` to `*-label`](https://github.com/asciidoctor/asciidoctor/issues/702) but this should not prevent to document the current state in the current doc. I guess the renaming issue will not be resolved soon. We need to document what exist today.

As for the beta status of the PDF rendering... It is already so good that I did not notice its beta status.

For me the user-manual page should mention every existing thing (I use CTRL-F a lot in this docucument).

mojavelinux:
The beta status isn't about quality or completeness, but rather about whether the APIs are subject to change. They are very much still in flux (and there's no test suite to validate those changes), hence why it is still alpha.

mojavelinux:
Since this is such an outlier, I decided to take your advice and include it in the user manual. I made some small revisions when I conducted the merge. Feedback welcome.

Merged as 1ae9935f6d0da5adadad871fe5f5caacc6de6ee3.

Thanks!

Issue #433, opened 2015-12-11
"Introduction To Asciidoctor" Content Discussion


jaredmorgs:
http://asciidoctor.org/docs/user-manual/#introduction-to-asciidoctor is currently available to create content for. 

Use this issue as a tracker for content suggestions, or feel free to provide PRs referencing this ticket to keep track of contributions.

https://github.com/asciidoctor/asciidoctor.org/blob/master/docs/user-manual.adoc is where this heading is located.

mojavelinux:
At last, there is an introduction page for Asciidoctor. See https://docs.asciidoctor.org/asciidoctor/latest/

PR #587, opened 2016-05-25
Abbreviations should be written out on their first use (#433)

sboekhoff:


mojavelinux:
Thanks @sboekhoff! There's only one problem which your change has surfaced. The Asciidoctor processor no longer uses ERB templates internally. We should simply say that it provides built-in converters. The first sentence of the following section contains the correct explanation.

https://github.com/asciidoctor/asciidoctor/#asciidoc-processing

sboekhoff:
OK, I think this problem deserves a larger review.

PR #590, opened 2016-05-27
Quickref fixes

sboekhoff:
Some of the results did not match the expected results for the shown code.

### DIFF
--- a/docs/_includes/ex-header-attr.adoc
+++ b/docs/_includes/ex-header-attr.adoc
@@ -20,7 +20,7 @@ My document provides...
 // tag::b-attr[]
 :homepage: http://asciidoctor.org
 :docslink: http://asciidoctor.org/docs[Asciidoctor's Docs]
-:description: Asciidoctor is a mature, plain-text document format for +
+:summary: Asciidoctor is a mature, plain-text document format for +
        writing notes, articles, documentation, books, and more. +
        It's also a text processor & toolchain for translating +
        documents into various output formats (i.e., backends), +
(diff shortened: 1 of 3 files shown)


docs/_includes/ex-header-attr.adoc
```diff
@@ -20,7 +20,7 @@ My document provides...
 // tag::b-attr[]
 :homepage: http://asciidoctor.org
 :docslink: http://asciidoctor.org/docs[Asciidoctor's Docs]
-:description: Asciidoctor is a mature, plain-text document format for +
+:summary: Asciidoctor is a mature, plain-text document format for +
```
mojavelinux: The choice to use `description` as the attribute name is important because it gets mapped to the description meta element in the HTML output.

mojavelinux:
If you revert the attribute name change, I'll merge. Thanks!

sboekhoff:
@mojavelinux : But then the result shown does not match the expected result for the code of the example.

att_underline

Do you suggest that I change the value of `description` in the example to the value used in `asciidoc-syntax-quick-reference.adoc` instead? Would be fine for me. But the next time the value changes in `asciidoc-syntax-quick-reference.adoc` it also needs to be changed in `ex-header-attr.adoc`.

mojavelinux:
It likes we are getting a side effect. We need to get that fixed so that they match.

mojavelinux:
After looking closer at this, I realize that it is not possible to redefine an attribute in an AsciiDoc table cell that is defined in the parent document. We use an AsciiDoc table cell to display the result. Therefore, I'm okay with this change...though in general this example is fragile for this reason.

mojavelinux:
Thanks!

PR #592, opened 2016-05-30
Fixed result of listing block example showing wrong result.

lucascosti:
See http://asciidoctor.org/docs/user-manual/#listing-blocks, example 3 for the erroneous result.


mojavelinux:
Thanks @lucascosti!

PR #593, opened 2016-06-01
Typo in user manual

sboekhoff:


mojavelinux:
Thanks @sboekhoff! I love the needle in the haystack fixes the best. So hard to find!

PR #610, opened 2016-10-02
Remove duplicated templateDir property

meistermeier:
templateDir is described twice in the documentation. Both descriptions are right but it should be more consistent with the rest of the documentation.


mojavelinux:
Can you make this change in the canonical file at https://github.com/asciidoctor/asciidoctor-maven-plugin/blob/master/README.adoc? We'll then need to clear out the contents of this file and replace it with an include of that file (like we do with docs/asciidoctor-diagram.adoc).

meistermeier:
After reading the linked README I will close this pull request since the changes were already made.

PR #611, opened 2016-10-02
consistent property description for templateEngine

meistermeier:
to get a consistent configuration property description the default value for templateEngine should also be mentioned.


mojavelinux:
Can you make this change in the canonical file at https://github.com/asciidoctor/asciidoctor-maven-plugin/blob/master/README.adoc? We'll then need to clear out the contents of this file and replace it with an include of that file (like we do with docs/asciidoctor-diagram.adoc).

meistermeier:
After reading the linked README I will close this pull request since the changes were already made.

mojavelinux:
Great. I'll open a separate issue to change the document so it includes the content from the upstream repository instead of duplicating it. See #612.

PR #616, opened 2016-11-01
resolves #609 document that start attribute must be a number

### LINKED ISSUE #609: The start index of a list needs be set numeric
Let's suposse I want to start/continue a list with the letter `m`:

```
[start="13"]
.. Something here.


[start="m"]
m. Something here. (I get `a.` instead of `m.` and a WARNING: list item index: expected a, got m)   

[start="13"]
m. Something here. (WARNING: list item index: expected a, got m)


[start="13"]
.. Something here. (works!)
```

The result:

list-m

So my suggestion is to say in the manual that the start index of the list must be a numeric entry, even if we want a list with letters.

mojavelinux:


mojavelinux:
cc: @edusantana

edusantana:
:+1:

Issue #612, opened 2016-10-03
Include documentation for Maven plugin from upstream repository


mojavelinux:
Instead of maintaining a separate document for the Maven plugin, include the content from the upstream repository as is done with the Asciidoctor Diagram page.

This document:

https://github.com/asciidoctor/asciidoctor.org/blob/master/docs/asciidoctor-maven-plugin.adoc

should include this document:

https://github.com/asciidoctor/asciidoctor-maven-plugin/blob/master/README.adoc

However, we should link to a tag.

Before this can be done, outstanding differences need to be reconciled. The focus is to make sure no content is missing from the upstream that was added to the existing document.

mojavelinux:
See https://docs.asciidoctor.org/maven-tools/latest/

PR #619, opened 2016-11-22
bump asciidoctor-maven-plugin version to 1.5.3

NicolasGeraud:


mojavelinux:
Thanks. I'm going to merge this just so that we can get the information fixed, but what really needs to happen is that the website needs to include the README from the code repository so that we don't have two separate documents with the same content. The reason I haven't done that yet is because the two documents have diverged and need to first be reconciled.

NicolasGeraud:
Hi Dan,
They diverge from the way they organize content indeed, but the README looks more exhaustive.

mojavelinux:
Yes, I should have clarified. The README is now the canonical source, but is missing a few bits of information from the version on asciidoctor.org. Unfortunately, the only way to fix it is a full scan to see what is missing from the README. As soon as we get that content moved over, we can just include the README and put this behind us.

mojavelinux:
Btw, that's #612.

PR #620, opened 2016-11-23
Link to asciidoctor-reveal.js README instead of out-of-date doc

obilodeau:
Linking to the up-to-date README of the asciidoctor-reveal.js repository instead of out-dated documentation.

Decided to do that because Google sent me there and I think this is confusing and misleading to potential users.


mojavelinux:
Actually, what we need to do is include the README using the include directive like we do with the other modules. See https://github.com/asciidoctor/asciidoctor.org/blob/master/docs/asciidoctor-diagram.adoc.

I'll go ahead and set this up.

One thing that is problematic is that the page on asciidoctor.org doesn't have a link to the repository on GitHub. I'll also fix that.

Eventually, these are going to become proper project landing pages on asciidoctor.org.

mojavelinux:
I have set this up, which supersedes this PR. The new page is available at http://asciidoctor.org/docs/asciidoctor-revealjs/.

The page pulls the README from a release tag. To update the page on asciidoctor.org, all you need to do is send a PR to update the corresponding attribute in the document header. See

https://github.com/asciidoctor/asciidoctor.org/blob/master/docs/asciidoctor-revealjs.adoc

Note, however, that there are some hacks in this document currently until we get settings synchronized with the upstream README.

mojavelinux:
I also turned the GitHub icon into a link to the project repository so that Google can figure out the relationship. (We could probably still stand to do this better in the future).

obilodeau:
Nice stuff! I'm amazed by the composability of AsciiDoc.

mojavelinux:
:+1: 

(and we can do even better than that).

PR #621, opened 2016-11-25
Resolves #204 error message list

### LINKED ISSUE #204: Add section to user manual about log messages
Add a section to the user manual that explains the log messages, what the levels (e.g., ERROR, WARNING) mean and how to address them.

rockyallen:
Please don't accept this! I just want to check the layout with you before spending time populating it.

### DIFF
--- a/docs/_includes/messages.adoc
+++ b/docs/_includes/messages.adoc
@@ -1,0 +1,275 @@
+//== Application Messages
+
+// Only includes the CLI. asciidoctorj and the asciidoctor API are not included.
+// When there are enough documented, maybe add another table for them. Although these users are programmers-wouldn't they just run it in the debugger?
+
+////
+Testing was carried out with
+Asciidoctor 1.5.5 [http://asciidoctor.org]
+Runtime Environment (ruby 2.3.1p112 (2016-04-26) [i386-linux-gnu]) (lc:UTF-8 fs:UTF-8 in:- ex:UTF-8)
+////
+
+All warning, error, and failure messages emitted by the Asciidoctor CLI (i.e., `asciidoctor`) are listed in the table below.
+These messages are written to the console via stderr (i.e., standard error).
+Most messages also apply to the API, though keep in mind they're written directly to stderr.
+
+//.Legend
+//--
+.Placeholders
+//[horizontal]
+<docname>;; represents the basename of the source file being processed (e.g., `sample.adoc`).
+<file>;; represents a path to the input file or other referenced file.
+<uri>;; represents a URI being referenced.
+<x> or <y>;; placeholders for other contextual information in the message.
+
+.Categories
+//[horizontal]
+ERROR;; Errors do not stop conversion, but the output document will almost certainly be wrong.
+FAILURE;; Failures are fatal; no output document will be produced.
+WARNING;; Warnings do not stop conversion, but they indicate possible problems, and the output may not be what you were expecting.
+//--
+
+.List of Application Messages
+[options="header,breakable",cols="<20,<60,<50,<30"]
+|====
+|Category |Message |Cause |See also
+
+|ERROR
+|input file <file> missing or cannot be read
+|Check that the file exists and that the filename is not misspelled.
+|<<using-the-command-line-interface>>
+
+|ERROR
+|include file has illegal reference to ancestor of jail, auto-recovering
+|The safe mode is restricting access to an include file outside of the base directory.
+|<<running-asciidoctor-securely>>
+
+|ERROR
+|input file and output file cannot be the same: <file>
+|Choose a different output directory or filename.
+|
+
+|ERROR
+|partintro block can only be used when doctype is book and it\'s a child of a part section. Excluding block content.
+|Invalid book document structure.
+|<<book-parts-and-chapters>>
+
+|ERROR
+|unmatched macro: endif::<x>[]
+|`endif::[]` with no unclosed preceding `ifdef::<x>[]`.
+|<<ifdef-directive>>
+
+|ERROR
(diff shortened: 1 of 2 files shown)


mojavelinux:
I do like it, though I wonder if the category of the message should be a separate column. I don't think the "asciidoctor" prefix is necessary on the message since that is always the same.

rockyallen:
Table updated - comments added in italics where I don't understand.

docs/_includes/errormessages.adoc
```diff
@@ -0,0 +1,264 @@
...
+|<<using-the-command-line-interface>>
+
+|ERROR
+|include file has illegal reference to ancestor of jail, auto-recovering
+|
+|<<running-asciidoctor-securely>>
+
+|ERROR
+|IOError, %(input file and output file cannot be the same: #outfile)
+|__I cant produce this message. asciidoctor -o file.adoc file.adoc overwrites file.adoc silently.__
```
mojavelinux: We're missing a check for that case. I'll need to fix that. The case currently checked is:

```
asciidoctor -a outfilesuffix=.adoc file.adoc
```

docs/_includes/errormessages.adoc
```diff
@@ -0,0 +1,264 @@
...
+|
+
+|ERROR
+|IOError, %(target directory does not exist: #{to_dir})
+|__I cant produce this message. asciidoctor -D doesnotexist file.adoc creates doesnotexist silently.__
```
mojavelinux: This only happens when using Asciidoctor via the API, the mkdirs option is not set, and the target directory does not already exist.

docs/_includes/errormessages.adoc
```diff
@@ -0,0 +1,264 @@
...
+====
+
+.Common Error Messages
+[options="header,breakable",cols="<20,<60,<50,<30"]
+|====
+|Category (1)|Message |Cause |See also
+
+|ERROR
+|input file <file> missing or cannot be read
+|Misspelled filename?
```
mojavelinux: Let's write this as "check that the file exists and that the filename is not misspelled"

docs/_includes/errormessages.adoc
```diff
@@ -0,0 +1,264 @@
...
+|<<include-uri>>
+
+|WARNING
+|could not retrieve contents of <x> at URI: <uri>
+|Web address not found.
+|
+
+|WARNING
+|could not retrieve image data from URI: <uri>
+|__cant reproduce, even with :data-uri: and non existent file__
```
mojavelinux: Given the following AsciiDoc:

```
image::http://example.org/invalid-image.jpg[]
```

You can reproduce by running:

```
asciidoctor -a allow-uri-read -a data-uri sample.adoc
```

mojavelinux:
Looking good! In some cases, the error message still has some Ruby code in it. Can you strip those down to just the message? (look for trailing if statements or IOError() wrappers).

rockyallen:
Comments addressed, extra failure added, discovered  when I typed -b hhtml

docs/_includes/errormessages.adoc
```diff
@@ -0,0 +1,274 @@
...
+
+Errors do not stop conversion, but the output document will almost certainly be wrong.
+
+Warnings do not stop conversion, but they indicate possible problems, and the output may not be what you were expecting.
+
+// Rare/unexplained
+
+////
+|WARNING
+|gem \'thread_safe\' is not installed. This gem is recommended when registering custom converters.
```
mojavelinux: This warning occurs if you have registered a custom converter implementation and you try to use asciidoctor without first installing the thread_safe gem.

docs/_includes/errormessages.adoc
```diff
@@ -0,0 +1,274 @@
...
+|
+|
+
+|WARNING
+|gem \'thread_safe\' is not installed. This gem is recommended when using custom backend templates.
```
mojavelinux: This warning occurs if you use custom templates (`-T <template_dir>`) and you try to use asciidoctor without first installing the thread_safe gem.

docs/_includes/errormessages.adoc
```diff
@@ -0,0 +1,274 @@
...
+|
+|
+
+|WARNING
+|file does not exist or cannot be read: #{path}) if opts[:warn_on_failure]
+|__not tested. I dont know when it could occur, apart from include which is already covered__
```
mojavelinux: This happens if you specify `-a stylesheet=<file>` and `<file>` does not exist or is not readable (no permission).

docs/_includes/errormessages.adoc
```diff
@@ -0,0 +1,274 @@
...
+// use <x,y,z> for anything else
+
+////
+Testing was carried out with
+Asciidoctor 1.5.5 [http://asciidoctor.org]
+Runtime Environment (ruby 2.3.1p112 (2016-04-26) [i386-linux-gnu]) (lc:UTF-8 fs:UTF-8 in:- ex:UTF-8)
+////
+
+The most common error and warning messages when using the asciidoctor command line (including asciidoctor-ant) are listed alphabetically in the table below.
+Messages from asciidoctorj or the asciidoctor API are not included.
```
mojavelinux: Add a comment to this line about asciidoctorj and asciidoctor API. I think that just confuses the reader for now, but it is important information for the documenter.

mojavelinux:
A few more comments inline. I think we're almost to a first version.

rockyallen:
Updated as comments

mojavelinux:
I've made some revisions and rebased the PR into a single commit.

mojavelinux:
Let's go for it! Thanks!

PR #623, opened 2016-11-27
resolves #418 document preamble toc option

### LINKED ISSUE #418: Document the preamble toc option
The preamble toc option is mentioned, but never explained. It should have a section just like the macro option.

rockyallen:
Resolves #418 document toc=preamble.

Went further than I planned after experimenting with pdf and docbook:
* docbook is unaffected by toc, unless you are talking about asciictor-fopub which (might?) implement <?asciidctor-toc ?>
* left and right don't work in pdf


mojavelinux:
> docbook is unaffected by toc, unless you are talking about asciictor-fopub which (might?) implement

That's correct. The toc in DocBook is a switch. It's either on or off. The value is ignored. The DocBook toolchain doesn't have a way to control the toc location from DocBook (only via the XSL templates), so that's the best we can do given what we have.

> left and right don't work in pdf

Correct, nor does it make sense. At the moment, the toc is also a switch in Asciidoctor PDF. The converter cannot yet move it from the default location, which is after the title page (without extending the converter). For developments on this capability, see https://github.com/asciidoctor/asciidoctor-pdf/issues/233.

docs/_includes/migrating-from-asciidoc-python.adoc
```diff
@@ -196,11 +196,12 @@ No {adr} equivalent.
...
 // Ref migration guide
 |`toc2`
-.2+|`toc`
-.2+|Combined in a single attribute, see <<user-toc>>.
+.3+|`toc`
+.3+|Combined in a single attribute, see <<user-toc>>.
 
 // Ref migration guide
 |`toc-placement`
+|toc-position
```
mojavelinux: Can you format toc-position in monospace?

docs/_includes/toc-intro.adoc
```diff
@@ -5,11 +5,13 @@ Included in:
...
 A table of contents (TOC) is a list of section and nested subsection titles that is generated when Asciidoctor processes a document.
-The TOC is inserted directly below the document's title and author, titled _Table of Contents_, and lists the section 1 and section 2 level titles by default.
 
-When the `docbook` converter is used, a TOC is inserted by default.
-To include a TOC when using the `html` or a custom backend, you must set the `toc` attribute.
-The toc attribute can be specified via the command line (`-a toc`) or in the document's header (`:toc:`)
+To include a TOC, you must set the `toc` attribute.
+
+[NOTE]
+This section applies to the html and pdf backends only. `toc` is ignored by the docbook backend because the TOC in docbook is controlled by the docbook stylesheets.
```
mojavelinux: The `toc` is not completely ignored by the DocBook converter. It's just a switch. It is either set or not set. The placement is what is ignored.

docs/_includes/toc-layout.adoc
```diff
@@ -18,17 +18,19 @@ include::toc-left.adoc[]
...
 // end::column[]
 
-// How does the toc-position attribute fit into this section?
-
 === Manual placement
 // tag::place[]
-The `macro` value on the toc attribute allows you to place the TOC anywhere in your document.
-To control where the TOC is rendered in your document, set the toc attribute to `macro`.
-Then, within the body of your document, use the `+toc::[]+` block macro to indicate the TOC's location.
+`toc` values of `auto`, `preamble`, and `macro` place the TOC in the body of the document.
```
mojavelinux: We should say "allow the TOC to appear in the main document area". Technically, the toc is always in the body of the document. We just move it around using the stylesheet. But that doesn't really matter to the user. It's all about how it appears.

docs/_includes/toc-layout.adoc
```diff
@@ -18,17 +18,19 @@ include::toc-left.adoc[]
...
+
+When `toc` is blank or set to `auto`, the TOC is inserted directly below the document’s title and author.
+
+When `toc` is set to `preamble`, the TOC is placed immediately below the preamble.
+It will not appear if your document does not have a preamble.
+
+To place the TOC anywhere else in your document, set the `toc` attribute to `macro`.
+Then, put the `+toc::[]+` block macro where you want the TOC to appear.
```
mojavelinux: Mention that you can only use the `toc::[]` macro once in your document. If the `toc::[]` macro is present, but the value of the `toc` attribute is not `macro`, the macro will be ignored (so it's safe to leave it there).

docs/_includes/sum-toc.adoc
```diff
@@ -4,33 +4,41 @@ Included in:
...
-.3+|:toc:
-
+.4+|toc
+|auto, left, right, macro, preamble
+|`:toc: left`
 |Defaults to `auto` if value is blank.
 |html
 
-|_not set_, `macro`, and `preamble` only
+|macro, preamble
```
mojavelinux: Since 1.5.3, the auto location now works with embeddable HTML. I'll go ahead and add that in.

mojavelinux:
Merged as 746f8f36b92bd18986065cde08cdd209b49dd299.

Thanks!

PR #627, opened 2016-12-01
Resolves #381 image sizing

### LINKED ISSUE #381: Document the unit of measure for height and width of images
Document what unit of measure is used for height and width of images. The answer is CSS pixels, the same as defined in HTML5.

It would also be good to document how this affects PDF generation. Basically, the scaledwidth attribute can be used to scale the image to a certain width of the PDF page. Otherwise, it is up to the PDF conversion code to interpret how to translate CSS pixels into an appropriate size in the PDF.

rockyallen:
Resolves #381 image sizing

### DIFF
--- a/docs/_includes/imagesizing.adoc
+++ b/docs/_includes/imagesizing.adoc
@@ -1,0 +1,125 @@
+//= Sizing an image
+
+To control the image size, you should provide a separate attribute for each output:
+
+----
+image::flower.jpg[width="300",pdfwidth="50%",scaledwidth="50%"]
+----
+
+NOTE: Images always keep their original aspect ratio, so it is only necessary to specify the width.
+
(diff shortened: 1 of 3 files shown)


mojavelinux:
I made some revisions and pushed it as 19c76be06f3c02a0f4078f77d08bc3f13061b76e.

The main change I made was to clarify that width and height are really intended to be backend agnostic. They are informational attributes that communicate the intrinsic (or pseudo-intrinsic) size of an image. They also happen to be mapped to attributes of the same name in the HTML output. The other attributes (pdfwidth, scaledwidth and scale) are intended to tune the size of the image for other mediums. Eventually, I think we're going to want htmlwidth as well so that we can tune the size of the image for HTML independently from the intrinsic size of the image. But we're still exploring that idea.

PR #631, opened 2016-12-02
Resolves #630 variables

### LINKED ISSUE #630: "Variable" should be attribute in attribute section
http://asciidoctor.org/docs/user-manual/#assigning-document-variables-inline
is currently 

"13.6. Assigning Document Variables Inline
Document variables can be assigned using the following syntax:
{set:<attrname>[!][:<value>]}
This is important for being able to assign document attributes in places where attribute entry lines are not normally processed, such as in a table cell."

I assume this is an error rather than an attempt to define another attribute class.

rockyallen:
Resolves #630 variables

### DIFF
--- a/docs/user-manual.adoc
+++ b/docs/user-manual.adoc
@@ -483,9 +483,9 @@ property 1:: does stuff
 property 2:: does different stuff
 ----
 
-=== Assigning Document Variables Inline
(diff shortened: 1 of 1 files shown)


mojavelinux:
Thanks!

PR #632, opened 2016-12-02
Resolves #558 element attributes

### LINKED ISSUE #558: Clarify difference between document attributes and element attributes
...there's still some confusion between document attributes and element attributes. We should make this distinction very clear.

In general, Asciidoctor uses attributes as a way to store and deliver metadata. Document attributes are available everywhere and are used to control behavior of the document (they have a broad scope). Element attributes are specifically for controlling an individual element and are scoped to that element. Unlike document attributes, element attributes cannot be referenced using an attribute reference (though, that could change in UniDoc).

Element here refers to a node or a directive. A node is a paragraph, a delimited block, a block macro, an inline macro or a span of formatted text. (maybe that needs to go in the glossary).

rockyallen:
Resolves #558 element attributes

The brief was to clarify the difference between document and element attributes, but I got carried away.
Mainly I have tried to separate the syntax of role, options and id  from the purpose of the attributes.

### DIFF
--- a/docs/_includes/attr-element.adoc
+++ b/docs/_includes/attr-element.adoc
@@ -4,7 +4,151 @@ Included in:
 - user-manual: Attributes: Setting attributes on an element
 ////
 
+
+////
+
+...there's still some confusion between document attributes and element attributes. We should make this distinction very clear.
+
+In general, Asciidoctor uses attributes as a way to store and deliver metadata. Document attributes are available everywhere and are used to control behavior of the document (they have a broad scope). Element attributes are specifically for controlling an individual element and are scoped to that element. Unlike document attributes, element attributes cannot be referenced using an attribute reference (though, that could change in UniDoc).
+
+Element here refers to a node or a directive. A node is a paragraph, a delimited block, a block macro, an inline macro or a span of formatted text. (maybe that needs to go in the glossary).
+////
+
+// tag::revised[]
+
+
+Element attributes are defined in "attribute lists" between square brackets, and apply only to the immediately following "element",
+where an "element" can be:
+
+* a paragraph
+* a delimited block
+* a block macro
+* an inline macro
+* a span of formatted text
+
+The syntax for setting element attributes is different from that for document attributes, and there is shorthand available for the most common attributes.
+
+For example, in:
+
+----
+[WARNING]
+Keep hands away from moving machinery
+----
+
+* The attribute list contains a single attribute `style` (a "positional attribute"), which has the value `WARNING`. 
+
+* The element that it applies to is the single paragraph "Keep hands away from moving machinery".
+
+Element attributes override document attributes, so that if the example had been preceded by 
+
+----
+:style: quote
+----
+
+The style of the paragraph would still have been WARNING, 
+and the value of the `style` attribute before and after the paragraph would still be `quote`.
+
+==== Element attribute syntax
+
+Element attributes can be set using:
+
+* named attributes
+
+* shorthand
+
+* positional attributes
+
+Named attributes are simple name-value pairs:
+
+----
+[width="100m"]
+----
+
+The quotes are not necessary for simple strings.
+To unset an attribute (because there is a document attribute already set that you do not want), set its value to `none`.
+
+Because they are used so often, shorthand exists for setting the attributes `role` (.), `options` (%), and `id` (#), so
+
+----
+[#dogs.summary.incremental%unbreakable%header]
+----
+
+is exactly the same as 
+
+----
+[id="dogs",role="summary,incremental",options="unbreakable,header"]
+----
+
+Finally, if you do not give it any other clues, the element will assign attributes based on the position of the value in the list (a "positional attribute"), for example:
+
+.Positional attribute examples
+|====
+|Element |Position 1 |Position 2 |Position 3
+
+|Blocks
+|style
+|-
+|-
+
+
+|Block macros
+|style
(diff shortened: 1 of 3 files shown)


docs/_includes/attr-element.adoc
```diff
@@ -4,7 +4,151 @@ Included in:
...
+
+In general, Asciidoctor uses attributes as a way to store and deliver metadata. Document attributes are available everywhere and are used to control behavior of the document (they have a broad scope). Element attributes are specifically for controlling an individual element and are scoped to that element. Unlike document attributes, element attributes cannot be referenced using an attribute reference (though, that could change in UniDoc).
+
+Element here refers to a node or a directive. A node is a paragraph, a delimited block, a block macro, an inline macro or a span of formatted text. (maybe that needs to go in the glossary).
+////
+
+// tag::revised[]
+
+
+Element attributes are defined in "attribute lists" between square brackets, and apply only to the immediately following "element",
```
mojavelinux: We should probably say "immediately adjacent" instead of "immediately following"

docs/_includes/attr-element.adoc
```diff
@@ -4,7 +4,151 @@ Included in:
...
+where an "element" can be:
+
+* a paragraph
+* a delimited block
+* a block macro
+* an inline macro
+* a span of formatted text
```
mojavelinux: We should probably add "preprocessor directive" to this list, though it's definitely a distant cousin to the others in this list.

docs/_includes/attr-element.adoc
```diff
@@ -4,7 +4,151 @@ Included in:
...
+----
+[WARNING]
+Keep hands away from moving machinery
+----
+
+* The attribute list contains a single attribute `style` (a "positional attribute"), which has the value `WARNING`. 
+
+* The element that it applies to is the single paragraph "Keep hands away from moving machinery".
+
+Element attributes override document attributes, so that if the example had been preceded by
```
mojavelinux: I don't think we should point this out. The reason is that I really don't like this inheritance...at least not how it is currently implemented. I'd rather people not know about it, so let's just keep it commented out.

docs/_includes/attr-element.adoc
```diff
@@ -4,7 +4,151 @@ Included in:
...
+The style of the paragraph would still have been WARNING, 
+and the value of the `style` attribute before and after the paragraph would still be `quote`.
+
+==== Element attribute syntax
+
+Element attributes can be set using:
+
+* named attributes
+
+* shorthand
```
mojavelinux: Place "shorthand" third in the list.

docs/_includes/attr-element.adoc
```diff
@@ -4,7 +4,151 @@ Included in:
...
+
+* positional attributes
```
mojavelinux: I think that we should explain that the most common positional attribute is the name of a block, such as "quote". However, the recognized positional attributes are determined exclusively by the block, block macro or inline macro. For instance, a source block has three positional attributes: block name, language, line numbers flag (technically line numbers should be an option, but that's a separate discussion). The image macros have three positional attributes: alt text, width, height. At some point, we should probably document all of these. For now, it's enough to point out that they exist.

docs/_includes/attr-element.adoc
```diff
@@ -4,7 +4,151 @@ Included in:
...
 ////
 
+
+////
+
+...there's still some confusion between document attributes and element attributes. We should make this distinction very clear.
+
+In general, Asciidoctor uses attributes as a way to store and deliver metadata. Document attributes are available everywhere and are used to control behavior of the document (they have a broad scope). Element attributes are specifically for controlling an individual element and are scoped to that element. Unlike document attributes, element attributes cannot be referenced using an attribute reference (though, that could change in UniDoc).
+
+Element here refers to a node or a directive. A node is a paragraph, a delimited block, a block macro, an inline macro or a span of formatted text. (maybe that needs to go in the glossary).
```
mojavelinux: :+1:

docs/_includes/attr-element.adoc
```diff
@@ -4,7 +4,151 @@ Included in:
...
+The syntax for setting element attributes is different from that for document attributes, and there is shorthand available for the most common attributes.
+
+For example, in:
+
+----
+[WARNING]
+Keep hands away from moving machinery
+----
+
+* The attribute list contains a single attribute `style` (a "positional attribute"), which has the value `WARNING`.
```
mojavelinux: This attribute is also known as the block name. However, I think the right way to explain it is that the positional attribute is the style, and one of the aspects of the style is the block name (but may also include id, role and options). Maybe "style" isn't generic enough of a name for this group, though. Perhaps "signature"?

mojavelinux: In a way, it's kind of like a method signature.

mojavelinux:
I noticed that you commented out the Block Assignment and Inline Assignment sections. Do you feel like this information is now covered better elsewhere? I ask because I do think it's important still to explain that there are different ways to assign the id, roles, and options...and I don't want to lose that distinction.

docs/_includes/attr-element.adoc
olleolleolle: "Maybe note" inline - easy to overlook before being done.

mojavelinux:
I'm confident this information is now covered in the new docs site, mostly notably in this section: https://docs.asciidoctor.org/asciidoc/latest/key-concepts/#attributes

PR #635, opened 2016-12-05
Remove warning about extensions for v0.1.4

aih:
The reference to the status of Extensions in v0.1.4 is probably outdated. Any new information about extensions and their use should be added, and other references throughout the user manual should probably also be updated. Those are in example text, so they would not be substantive changes.


mojavelinux:
I agree it's time to drop this warning.

While the extension API is generally considered stable at this point, the one exception is the inline macro. Since it relies on the fact that the parsing of inline content is deferred to the conversion phase, it has to return converted text. This return value will change once we parse inline content during the parsing phase. So there will be some sort of migration (or adapter), but I have yet to determine what that will be.

mojavelinux:
Intro updated.

Thanks!

PR #637, opened 2016-12-08
Fix the table example with rolename

jmini:
As [discussed](http://discuss.asciidoctor.org/Table-amp-column-example-from-the-doc-is-not-working-td5129.html) on the mailing list, I have fixed the example "Table with rolename".

Because at this point of the docs, the `cols` attribute is not explained, I have changed the example with the rows defined on the same line (as in the previous example).

In `ex-table.adoc` file, I have moved the `tag::base[]` example (near to the other `base` examples). This snippet is used a lot in the `table-col.adoc` File.

### DIFF
--- a/docs/_includes/ex-table.adoc
+++ b/docs/_includes/ex-table.adoc
@@ -27,6 +27,29 @@ Examples for table sections
 |===
 // end::base-alt[]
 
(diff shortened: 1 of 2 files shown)


mojavelinux:
Thanks!

PR #642, opened 2016-12-30
Fix bower-based example, from the 2014-08-23 news

twitwi:
Even tough this news is quite old, it is still well ranked by search engines (for "asciidoctor bower"), so this fix might help some copy/paste beginners like me.


mojavelinux:
I'm all for making things right. Thanks!

PR #649, opened 2017-01-15
Resolves issue 648 List of FAQ

rockyallen:


docs/faq.adoc
```diff
@@ -1,4 +1,4 @@
-= FAQs and Troubleshooting
+= Frequently Asked Questions and Trouble Shooting
```
mojavelinux: I'd like to keep Troubleshooting as one word. Also, to help with search engines, let's write FAQs as:

Frequently Asked Questions (FAQs)

wdyt?

docs/faq.adoc
```diff
@@ -15,14 +15,29 @@ Tips for fixing installation, syntax, processing, and rendering problems when us
...
+:q3: What is the internet mime type (aka MIME type) for AsciiDoc?
+:q4: Part way through the document, the blocks stop rendering correctly. What went wrong?
+:q5: Why don't URLs containing underscores (`_`) or carets (`{caret}`) work after they're rendered?
+
+. xref:faq1[_{q1}_]
+. xref:faq1a[_{q1a}_]
+. xref:faq2[_{q2}_]
+. xref:faq3[_{q3}_]
+. xref:faq_rendering[_{q4}_]
+. xref:faq_urls[_{q5}_]
```
mojavelinux: I like this. Perhaps we should add a horizontal rule between the question list and the body. (Three hyphens).

rockyallen: Done

mojavelinux:
I decided to take a different approach with this. I used section titles instead. That way, we can leverage the built-in TOC. The more I use the qanda list, the more I'm unconvinced it's actually usable in the real world. Perhaps if it had a way to create an index or the questions were collapsible it would work. But I don't have time to mess with that right now.

mojavelinux:
See http://asciidoctor.org/docs/faq/.

PR #650, opened 2017-01-25
Add safe mode restriction note to documentation of linkcss

coy123:
When learning how to use asciidoctor, I had no idea that the safe mode SAFE is necessary to use the attribute linkcss. So I add it here to clarify for future users.
IMO, all attributes with such restrictions should clarify in their explanations.


mojavelinux:
You deleted the branch before the PR was merged. I'll apply manually.

mojavelinux:
I have documented this behavior.

PR #651, opened 2017-01-31
fix uri-image-attrs link

rlespinasse:
The link in the TIP from [**Inline image with positioning role**](http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#images) don't work correctly.


mojavelinux:
Thanks!

PR #653, opened 2017-03-08
Update table.adoc

tedbergeron:
Add word 'optional' to show a blank line between table rows is suggested, but not required.

### DIFF
--- a/docs/_includes/table.adoc
+++ b/docs/_includes/table.adoc
@@ -22,9 +22,9 @@ The simple table example below consists of two columns and three rows.
 include::ex-table.adoc[tags=base-co]
 ----
 <1> The table's content boundaries are defined by a vertical bar followed by three equal signs (`|===`).
-<2> A blank line separates the table delimiter and the first row so the first row is not styled as the table's header.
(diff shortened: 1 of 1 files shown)


tedbergeron:
I could not replicate that the blank line separating the table delimiter from the first row would force the row to be styled as the table header in Asciidoctor v 1.5.4. Perhaps this is no longer the default behavior.

[options="header"] forces row one to be rendered as <th> and adds classes.

mojavelinux:
Thanks for these suggestions. I merged them as 9feea14d05b8f8f1aea7c8e6f601ea1ce2a1ea56.

mojavelinux:
> I could not replicate that the blank line separating the table delimiter from the first row would force the row to be styled as the table header in Asciidoctor v 1.5.4.

Actually, it's the opposite logic. Inserting a blank line before the first row disables the automatic promotion of the first row to the table header. I clarified this statement in the revision.

The following table would have a header row:

```
|===
|A |B

|1 |2
|===
```

The following would not:

```
|===

|A |B

|1 |2
|===
```

mojavelinux:
But yes, it's true you can also use either the header or noheader option to control whether or not the table has a header row explicitly.

Issue #398, opened 2015-06-24
Document the outfilesuffix and how it's used for relative links on GitHub


mojavelinux:
See https://github.com/asciidoctor/asciidoctor/issues/417#issuecomment-36426214

DanielSWolf:
It seems to me that the linked issue has seen a number of revisions, so I'm not sure what the final syntax is. The [official documentation](https://github.com/asciidoctor/asciidoctor/wiki/How-to-link-to-a-relative-AsciiDoc-document) seems to be outdated. I'd appreciate it if you could find the time to document the this.

mojavelinux:
This is now documented here: https://asciidoctor.org/docs/user-manual/#navigating-between-source-files

If, after reading that, it's still not clear, please open a new issue so we can address it.

PR #655, opened 2017-03-15
#398: Update asciidoc-writers-guide.adoc

hemanik:
Added tip to include relative link to a file in a sub directory using xref inline macro.
issue:#398

### DIFF
--- a/docs/asciidoc-writers-guide.adoc
+++ b/docs/asciidoc-writers-guide.adoc
@@ -366,6 +366,8 @@ include::{includedir}/xref.adoc[]
 
 include::{includedir}/xref-relative.adoc[]
 
+TIP: The xref inline macro by default resolves the link to HTML, to create a relative link to an asciidoc file in another subdirectory include `ifdef::env-github[:outfilesuffix: .adoc]` below the document header and drop the ".adoc" extension after the filename.
+
 Image references are similar to links since they are also references to URLs or files.
 The difference, of course, is that they display the image in the document.
 


docs/asciidoc-writers-guide.adoc
```diff
@@ -366,6 +366,8 @@ include::{includedir}/xref.adoc[]
 
 include::{includedir}/xref-relative.adoc[]
 
+TIP: The xref inline macro by default resolves the link to HTML, to create a relative link to an asciidoc file in another subdirectory include `ifdef::env-github[:outfilesuffix: .adoc]` below the document header and drop the ".adoc" extension after the filename.
```
ecki: Is there a method which is more neutral (support any output or source format). I canimagine there are more edit environents than GitHub?

hemanik: @ecki, Thanks for pointing out, it shouldn't indeed be limited to just GitHub. 
I am not aware of any generic way, will explore and update the PR.

mojavelinux: It's important to note that the xref macro only links to another file if the target ends in a hash path (e.g., `#` or `#anchor-name`). The `#` is how Asciidoctor distinguishes between an internal xref and an external xref.

> Is there a method which is more neutral (support any output or source format).

The `env-*` attributes are set by the integrations. For example, the code that runs on GitHub (and GitLab) sets the `env-github` attribute. You should expect that each integration has a similar attribute. For example, Awestruct sets `env-site`. We've tried to introduce generic attributes where possible, but in some cases like GitHub, the environment is so unique that it really only applies to that specific integration.

Btw, the browser previews also need the `outfilesuffix` to be `.adoc`. Both the Chrome and Firefox plugin set this value by default now. You can also use `env-browser` in those environments.

mojavelinux: Eventually, we'll want to document all these environment attributes in the user manual. At the time of this writing, they are still evolving.

mojavelinux: I'd like to see outfilesuffix set to `.adoc` by default on GitHub. There's an issue tracking that change. See https://github.com/github/markup/issues/519

mojavelinux:
I incorporated this into https://github.com/asciidoctor/asciidoctor.org/commit/731a2878f40d976db9c2997bcf536034b1eb9da1.

This is now the default behavior on GitHub.

PR #657, opened 2017-03-19
README: Linked dependencies

ecki:
For Foundation it might be imprtat to note it is the old version.

The Travis CI link is optional as the is a build batch next to it. I refrained from using {:Awestruct} logo text, but that would be another possible addition.


ecki:
Superceeded by PR#658

PR #658, opened 2017-03-19
README: link to external frameworks used

ecki:
Make it explicite that older Awestruc is used.


mojavelinux:
I think you mean older Foundation (based on the link you used).

mojavelinux:
Thanks!

PR #659, opened 2017-03-23
Update revnumber to include the latest readme

madmas:
Update readme to include the latest readme for asciidoctor-diagram. I was missing the diagram attributes table recently and think it should be accessible from a more popular location ;-)


mojavelinux:
Yes, this is the idea. When Asciidoctor Diagram is released, this number should be updated. We plan to automate this in the future.

mojavelinux:
Thanks!

PR #669, opened 2017-05-09
Update docs to inform about :smartquotes:

giflw:
The option :smartquotes: was undocumented


mojavelinux:
Um. That's because it's not implemented. Right?

giflw:
Sorry, I made something wrong here and when I used :smartquotes: it begin to work. Tested again without it and its working.

mojavelinux:
:wink:

PR #671, opened 2017-05-19
Resolves #670 counters must be lower case

### LINKED ISSUE #670: Document that counters must be lower case
See asciidoctor bug #2123

rockyallen:
Resolve #670


docs/_includes/counter.adoc
```diff
@@ -12,7 +12,7 @@ Possible uses include inline itemized lists, paragraph numbers and serial item n
 
 A counter is implemented as a specialized document attribute.
 You declare and display a counter using an augmented attribute reference, in which the attribute name is prefixed with `counter:` (e.g., `+{counter:name}+`).
-Since counters are attributes, counter names follow the same rules as attribute names.
+Counter names follow the same rules as <<attribute-restrictions,attribute names>> *except* that any letters must be lower case.
```
mojavelinux: I'd go so far as to say all attribute names should be lowercase. While the attribute entries (the definition) may contain uppercase characters, those characters are lower-cased before the attribute is saved. So technically, it's only allowed in style. And even then it shouldn't be done. That's how I think we should document this.

mojavelinux:
I merged as 5a73776aab6b5a1d3df2b5994b912380dcfcd281.

PR #673, opened 2017-05-19
Resolves #672 document list numbers

### LINKED ISSUE #672: Document table-number attribute etc
They are document attributes that automatically created, and used to number formal tables (figures, listings).
They should probably be left alone, although you can override them to them alter the sequence (or is this a hack to be warned against?). Either way they should be listed so you don't accidentally clobber them.
Closest category is "Environment" attributes.

rockyallen:
Resolves #672 document list numbers

### DIFF
--- a/docs/_includes/attrs-env.adoc
+++ b/docs/_includes/attrs-env.adoc
@@ -18,10 +18,6 @@ It's recommended that you treat these attributes as read only.
 |===
 |Attribute |Description |Example Value
 
-|appendix-number
-|The number of the current appendix^[<<note_reserved,3>>]^.
-|5
-
 |asciidoctor
 |Set if the current processor is Asciidoctor.
 |{asciidoctor}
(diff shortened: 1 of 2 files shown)


mojavelinux:
Generally, *-number attribute is generated for a block with a title only if there's a matching *-caption attribute defined. There are some special cases. Here's the complete list:

* example-number (since example-caption exists)
* table-number (since table-caption exists; in reality, it's not checking but should be)
* listing-number (only if listing-caption is defined)
* figure-number (a special case for block images; not checking for figure-caption, but should be)
* appendix-number (always set for an appendix section)

docs/_includes/attrs-env.adoc
```diff
@@ -67,6 +67,11 @@ For example, if the backend is `docbook5`, the basebackend is `docbook`.
 |Set if content is being converted to an embeddable document (body only).
 |
 
+|figure-number
+|The number of the most recent formal (ie titled) figure.
+You can set it to any integer value to change the subsequent figure numbers.
```
mojavelinux: Let's not state that this attribute can be changed. Instead, let's emphasize that it's auto-generated and should thus be treated as a reserved name.

mojavelinux:
Oops, I forgot about chapter-number.

rockyallen:
I think this is now ready

mojavelinux:
I decided to move these attributes to the built-in table since they aren't really environment attributes. I grouped them in the Internationalization section (renaming it to Internationalization and numbering) since they influence the caption in a way similar to those other attributes.

mojavelinux:
Thanks!

Let me know if you're happy with the final result. See http://asciidoctor.org/docs/user-manual/#builtin-attributes-i18n.

PR #676, opened 2017-05-22
Add missing . in ifeval examples

janicecmt:


mojavelinux:
Actually, the `outfilesuffix` already includes the leading period, so there's no need to add it explicitly.

See http://asciidoctor.org/docs/user-manual/#env-attributes-table (scroll down to entry for outfilesuffix).

janicecmt:
That's just bizarre and completely unintuitive.

BTW, I just assumed that was a random name you made up to match an example where it makes sense to have multiple variables (I don't know all of the supported default attributes off the top of my head and doubt most people do)- it might make sense to add a note to the example that the outfilesuffix starts with . or at least have a link to its definition. It just looks wrong as is, so adding some clarification would be helpful.

I do think it's a useful example to have {var1} <characters> {var2} to show that you can interact that way so perhaps adding one like that would also be nice.

mojavelinux:
I've updated the docs to draw attention to this fact.

The outfilesuffix value is consistent with the return value of Ruby's File.extname and Python's os.path.splitext, which both include the leading period. It's also necessary for compliance with AsciiDoc Python. But more than that, it's important to include the leading period so it can be used without condition to drop the file extension or append a URL segment.

mojavelinux:
> I don't know all of the supported default attributes off the top of my head and doubt most people do

That's why we made this handy table: http://asciidoctor.org/docs/user-manual/#attribute-catalog

janicecmt:
Yes, but I'm not going to look up the stuff used in every example. In any case, thanks for adding clarification.

I'm somewhat familiar with Ruby (I've looked at Ruby code before and can usually recognize it when I see it) but I've never really used it. I'm a bit more familiar with Python but it's still not a primary language and I clearly haven't done much file manipulation there :)

I've only seen or constructed cases where the period is separately specified - it provides clarity in my mind so that makes sense to me - and I've seen lots of them over the years, hence it being an expected norm in my brain. The Python, Ruby, etc choices are interesting; I can see what your saying about it making sure folks don't forget the period.

Anyway, thanks again.

mojavelinux:
I could see a case for storing just the name part in a separate attribute. I'm just unsure what to call it that wouldn't cause confusion. Perhaps outfileextension. Kind of long, but both "suffix" and "extname" are reserved to mean "include the leading period". Just an idea.

janicecmt:
I don't know how other folks think, but I vote for additional clarity over brevity. I think outfile-extension would be great.

PR #677, opened 2017-06-14
Spelling/grammar: "An optional revision, a date and an option remark"

dflock:
Correct `An optional revision, a date and an option remark` to `An optional revision, a date and an optional remark`, option -> optional.


mojavelinux:
Thanks! I love getting fixes like this. Every word counts.

PR #682, opened 2017-07-23
Resolves #681 tag expressions

### LINKED ISSUE #681: Document wildcards and negation operator when including by tags
!, *,. ** ref Asciidoctor issue 1516

rockyallen:


docs/_includes/include-lines-tags.adoc
```diff
@@ -32,9 +32,10 @@ include::ex-include.adoc[tags=tag-co]
 IMPORTANT: The `tag::[]` and `end::[]` directives should be placed after a line comment as defined by the language of the source file.
 The directives must also appear at the _end_ of the line.
 In the previous example, we choose to prefix the lines with a hash (`#`) because that's the start of a line comment in Ruby.
+For XML files you can use the `+<!-- tag::name[] -->+` and `+<!-- end::name[] -->+` delimiters.
```
mojavelinux: We should emphasize that the space before and after the directive is required.

docs/_includes/include-lines-tags.adoc
```diff
@@ -32,9 +32,10 @@ include::ex-include.adoc[tags=tag-co]
...
 
-// moved from include-directive.adoc, since I assume it refers to tagging. Should "targets" be tags?
-TIP: Asciidoctor supports targets that include spaces and the `\{sp}` attribute references.
+// moved from include-directive.adoc, since I assume it refers to tagging.
+TIP: Asciidoctor supports tags that include spaces and the `\{sp}` attribute references.
```
mojavelinux: "target" is the right term here. We are referring to the target of the include directive, otherwise known as the path being included. Spaces are permitted in the filename. As an alternative, the `{sp}` attribute reference is also recognized (though no longer preferred).

docs/_includes/include-lines-tags.adoc
```diff
@@ -88,9 +89,37 @@ snippet b
 
 Notice that none of the lines with the tag directives are displayed.
 
-NOTE: for XML files you can use the `+<!-- tag::name[] -->+` and `+<!-- end::name[] -->+` delimiters
+==== By expressions using tags
```
mojavelinux: I think we should name this section "Tag filtering". We are essentially getting more fine-grained about which tagged regions should be selected.

docs/_includes/include-lines-tags.adoc
```diff
@@ -88,9 +89,37 @@ snippet b
...
+ Non-tagged regions
+
+`!`::
+ Exclude
+
+The inclusive operations (explicit names, `\*` and `**`) are applied first, then the exclusive operations (`!`), so it doesn't matter which order they appear in.
+
+Some examples:
+
+`tags=a;!b` selects regions tagged a but excludes any _nested_ regions tagged b.
```
mojavelinux: Perhaps use "foo" and "bar" here instead of "a" and "b" so it's easier to see when you are referring to a tag name.

rockyallen:
As requested.

mojavelinux:
I made some revisions for accuracy. I also revised the intro, used the term tag macro instead of tag directive, added a dedicated tip about wrapping tags in circumflex comments, and applied some formatting and minor phrasing improvements.

PR #685, opened 2017-07-25
Resolves #684 tsv

### LINKED ISSUE #684: Document \t for table cell separators.
At the same time, clarify the difference between Asciidoctor processing and CSV processing.

I think:
- |=== or format=psv set Asciidoctor processing and the default separator = |
- ,=== or format=csv set CSV processing, and the default separator = ,
- format=dsv is the same as format=csv, separator=:
- format=tsv is the same as format=csv, separator=\t
- ,=== with format=psv feels like it should be an error, but is actually results in psv.

"CSV" processing is 1 row per line, separators _between_ fields, quoting using "", and no cell-level formatting. Cell content can be Asciidoc if it is in the source file or included:: from an Asciidoctor file type (but not any other extension). So you can have Asciidoc content in CSV files, you just need to rename them to .adoc.
Should also explain what it does If content is not strict RFC 4180, ie blank lines, extra spaces, ragged rows, and  embedded newlines in quoted values.

rockyallen:
Also reworked CSV linking to RFC 4180 and made data format more explicit.

### DIFF
--- a/docs/_includes/ex-table-data.adoc
+++ b/docs/_includes/ex-table-data.adoc
@@ -3,7 +3,7 @@ Examples for table data format section
 ////
 
 // tag::csv[]
-[format="csv", options="header"]
+[%header,format=csv]
 |===
 Artist,Track,Genre
 Baauer,Harlem Shake,Hip Hop
@@ -12,7 +12,7 @@ The Lumineers,Ho Hey,Folk Rock
 // end::csv[]
 
 // tag::dsv[]
-[format="dsv", options="header"]
+[%header,format=dsv]
 |===
 Artist:Track:Genre
 Robyn:Indestructable:Dance
--- a/docs/_includes/sum-table.adoc
+++ b/docs/_includes/sum-table.adoc
@@ -15,17 +15,20 @@ d|specifiers
 |see <<cols>> for details
 |
 
-.3+|format
-.3+|data format of the table's contents
+.4+|format
+.4+|data format of the table's contents
 |psv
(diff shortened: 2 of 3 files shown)


mojavelinux:
This is a great change.

I think we should title this section "Delimiter-separated values" and link to https://en.wikipedia.org/wiki/Delimiter-separated_values as the main reference. Comma-separated is just one of the three delimiter-separated formats supported. And it actually has slightly different parsing rules than TSV and DSV (though they are subtle).

In particular, DSV does not support newlines in data by quoting the value. Thus, the following doesn't combine the "c" and "d" into a single cell.

```
[format="dsv"]
|===
a:b:"c
d":e:f
|===
```

> Embedded newlines are allowed in the quoted string, but will be converted to single spaces.

We should mention this is actually a known bug. See https://github.com/asciidoctor/asciidoctor/issues/2041.

I also noticed that you promoted the prose that was interspersed between the examples and pushed the examples to an example section. Do you think this makes it easier for the reader? Or should we try to interleave the descriptions and the examples?

mojavelinux:
I made some revisions, rearranged some of the sections, made the introduction stronger, and drew specific attention to the TSV format. I also restored the reference to RFC 4180.

mojavelinux:
Here's the final result. If you have further suggestions, feel free to send a follow-up PR.

http://asciidoctor.org/docs/user-manual/#delimiter-separated-values

mojavelinux:
Thanks for your work on this @rockyallen!

PR #688, opened 2017-07-29
Clarify explanation for unconstrained example

### LINKED ISSUE #686: Examples for unconstrained formatting are contradictory
The section [When should I use unconstrained quotes?](https://github.com/asciidoctor/asciidoctor.org/blob/master/docs/_includes/formatting-marks.adoc#when-should-i-use-unconstrained-quotes) is not clear to me, especially the two examples

 * `–**2016**`: *Unconstrained, because of the “–” to the left of the emboldened number.*
  However, “–” is not a letter, number, underscore, colon or semi-colon. So according to the rules, unconstrained formatting should not be needed.
 * `*2016*–`: *Constrained, because the “&” is not a letter, number, underscore, or semi-colon.*
  There is no ampersand there. Maybe that's just a typo?

tribut:
This is the last remnant of my changes to fix #686 😩

### DIFF
--- a/docs/_includes/formatting-marks.adoc
+++ b/docs/_includes/formatting-marks.adoc
@@ -82,7 +82,7 @@ To help you determine whether a particular syntax pattern requires unconstrained
 |`+&ndash;**2016**+`
 |&ndash;**2016**
 |Unconstrained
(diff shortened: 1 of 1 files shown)


mojavelinux:
Good suggestions. Thanks!

PR #705, opened 2017-08-07
Xrefstyle full and short require "*-caption" attribute

jmini:


mojavelinux:
Thanks for adding this clarification!

PR #712, opened 2017-08-27
Resolves issue 711 expand positional attributes

rockyallen:
Use sentences


docs/_includes/attr-element.adoc
```diff
@@ -20,13 +20,16 @@ Attribute lists:
...
-in an attribute list
+Positional attributes are un-named values at the start of the attribute list.
+The attribute that they are assigned to depends on the type of the element:
 
-not named
+* `icon:` (1) size
+* `image:` and `image::` (1) alt text, (2) width, (3) height
+* Delimited blocks: (1) block name
+* Other inline quoted text: (1) `role`
+* Other blocks and macros: (1) `style`
```
mojavelinux: Technically block name and style are the same thing. The way I would write this is:

```
Delimited blocks: (1) block name (aka style)
```

Then just leave off the "other block and macros" since there is no universal positional attribute for those.

docs/_includes/attr-element.adoc
```diff
@@ -20,13 +20,16 @@ Attribute lists:
...
 
-the first positional attribute in the list on inline quoted text is referred to as the role attribute
-
-the first positional attribute in the list on blocks and macros is known as the style attribute
+For example, `image:[Sunset,300,400]` is equivalent to `image:[alt=Sunset, width=300, height=400]`.
```
mojavelinux: Please including a dummy image target since a blank image target is not a valid image macro.

mojavelinux:
Thanks!

PR #717, opened 2017-10-03
Update header-title.adoc

andrewcarver:
The listing entitled "Document with a subtitle using a custom separator" INCLUDEs the same tagged region (sub-2) as does the previous listing (entitled "Document with a subtitle and multiple colons"). Apparently this is just a typo, and it was intended to point rather to the subsequent tagged region (sub-3).


mojavelinux:
Thanks @andrewcarver!

PR #719, opened 2017-10-06
Update formatting-marks.adoc

andrewcarver:
Apply an attribute to `pass:[]`, not a passthrough.

Though it seems sensible on first blush, a passthrough will not permit a passthrough - to pass through (!). What's coming through instead is `<code></code>` . Though I don't know, I guess this is because within macros, macro-substitution takes place (sec. 40.5)...?

These are the only two places this exact AsciiDoc occurs within this file. I did find within the User-Doc HTML the `<code></code>` pattern in one other section - 46.2 - to where I'll now trot off and have a look :)

### DIFF
--- a/docs/_includes/formatting-marks.adoc
+++ b/docs/_includes/formatting-marks.adoc
@@ -199,7 +199,7 @@ It means to format TEXT as monospace, but don't interpolate formatting marks or 
 It's roughly equivalent to Markdown's backticks.
 Since AsciiDoc offers more advanced formatting, the double enclosure is necessary.
 
-The more brute-force solution to the inline passthrough approach is to use the `+pass:c[]+` macro, which is a more verbose (and flexible) version of the plus formatting marks.
+The more brute-force solution to the inline passthrough approach is to use the pass:q[`pass:c[\]`] macro, which is a more verbose (and flexible) version of the plus formatting marks.
 
 ----
 The pass:c[__kernel] qualifier can be used with the pass:c[__attribute__] keyword...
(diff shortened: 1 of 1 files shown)


mojavelinux:
Documenting AsciiDoc with AsciiDoc can sometimes be a tricky ordeal. This is one of those cases. But it goes beyond tricky in this case. You've uncovered a bug in the parser.

The following syntax should work:

```
`\pass:c[]`
```

However, the parser was processing the escaped macro on a second pass.

I've opened an issue to track this issue. See https://github.com/asciidoctor/asciidoctor/issues/2442

In order to have something that will work now and in 1.5.7 where this bug will be fixed, we have to use the more brute force syntax:

```
pass:q[`pass:c[\]`]
```

mojavelinux:
I updated the PR.

mojavelinux:
Thanks for the heads up!

andrewcarver:
You're welcome!
Please allow me to point out something further here I think is weird; it may help shed light on the bug:
Here is a list of 4 AsciiDoc snippets (in Normal text block), and what HTML I'm getting from them:

``pass:c[hello]``          -->   hello
``pass:c[h]``                -->    h
``pass:c[ ]``                 -->     
``pass:c[]``                  -->    pass:c[]

Now what the heck is with that last one? The only difference between it and the third, in the AsciiDoc, is the space between the square brackets is now lacking. The result surprises me -- but does it surprise you? (I hope that the formatting comes out correctly, I am hoping it shows that I'm surrounding each of these snippets with backticks in the AsciiDoc.)

andrewcarver:
Oh, I used AsciiDoc 1.5.5, ostensibly, for those 4 tests (packaged in asciidoctor-cli.js, running in node.js).

andrewcarver:
I meant, AsciiDoc**tor** 1.5.5 !!

mojavelinux:
This was a bug that existed before 1.5.6. Asciidoctor was not processing the inline pass macro with an empty value between the brackets. Now it does.

andrewcarver:
Cool!

PR #720, opened 2017-10-06
Apply style attribute to ++++ being taken as macro

andrewcarver:
Though it seems intuitive, putting `++++` inline in normal text doesn't work. What's coming through from it is `<code></code>`.

I'm guessing it is treated as a "single plus" passthrough macro, within another -- and, perhaps in keeping with sec. 40.5, the inner passthrough is processed (passing an empty value), and then the outer one (doing likewise!).

### DIFF
--- a/docs/_includes/stem.adoc
+++ b/docs/_includes/stem.adoc
@@ -72,7 +72,7 @@ Block formulas are marked up by assigning the `stem` style to a delimited passth
 include::ex-stem.adoc[tag=bl-co]
 ----
 <1> Assign the stem style to the passthrough block.
-<2> A passthrough block is delimited by a line of four consecutive plus signs (`++++`).
+<2> A passthrough block is delimited by a line of four consecutive plus signs (`pass:[++++]`).
 
 The result is rendered beautifully in the browser thanks to MathJax!
 


andrewcarver:
Attempt #2: I had to change my solution -- which I apologize for, I should have tested first!!
I guess I analyzed the problem wrong -- it's not being taken as a passthru within a passthru, else I think pass:[`++++`] would have worked... but it didn't either. 

Be that as it may, I'll applaud you if you can find a less verbose solution than this second edit. Good luck!

andrewcarver:
Or, you CAN use back-ticks instead of the `<code>` markup if you wrap the whole `pass:[++++]` in them.

mojavelinux:
The correct way to write this would be:

```
`pass:[++++]`
```

FYI: adding attributes to an inline phrase doesn't change how the formatting is interpreted. All that does is add a CSS class to the output.

There is one special role named x- that does enables legacy parsing, but we should phase out any usage of that in the user manual.

mojavelinux:
Thanks for catching this!

andrewcarver:
Thank you, Dan, for all the explanations!

mojavelinux:
My pleasure! Thanks for helping out!

PR #721, opened 2017-10-06
Remove redundant escape '\'s in literal columns

andrewcarver:
I don't see the point in having backslash escapes inside a literal column. 

I'm guessing that the escapes had already been entered, and later the author of the table decided to make columns 2 and 3 literal-format. But, the escapes "escaped" eviction (even tho they had become redundant).

### DIFF
--- a/docs/_includes/subs-symbol-repl.adoc
+++ b/docs/_includes/subs-symbol-repl.adoc
@@ -12,64 +12,64 @@ Included in:
 |Name |Syntax |Unicode Replacement |Rendered |Notes
 
 |Copyright
-|\(C)
-|\&#169;
+|(C)
+|&#169;
 |(C)
 |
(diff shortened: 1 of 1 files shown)


mojavelinux:
One of the bugs in Asciidoctor prior to 1.5.6 is that it would apply substitutions to contents of literal cells. That was fixed in 1.5.6. (See https://github.com/asciidoctor/asciidoctor/issues/1912).

So these are residual workarounds,

mojavelinux:
Thanks!

andrewcarver:
Ah yes. How you manage to keep even half of the documentation up to date with the bug fixes -- is something that hurts my brain to ponder.

mojavelinux:
All I can say is "many eyes" ;)

PR #722, opened 2017-10-06
Remove redundant \ escape in a literal-style block

andrewcarver:
I'm not sure how formatting is done in a man page -- I've written exactly zero of them.
However, I'm guessing that the literal styling this AsciiDoc source is shown in, makes that \ escape of the copyright symbol-code redundant.


mojavelinux:
This was actually intentional, though more like a habit. A lot of existing man pages use the literal (C) instead of &#169;. However, there are also man pages that use &#169;. To avoid confusion, we'll just go ahead and let it get interpreted in this example.

PR #723, opened 2017-10-06
Remove more redundant escapes in literal columns

andrewcarver:
Same problem here as in #721: here it is only HTML entities that are escaped, in a column that I'm guessing was only later styled as literal.


mojavelinux:
Thanks!

PR #724, opened 2017-10-06
Remove yet more escapes from literal HTML entities

andrewcarver:
I believe this marks the last of the backslash escapes that are themselves being inadvertently escaped by being in a literal column or block. (This is the same problem as described in #723 -- only occurring in Appendix A.3 instead.)


mojavelinux:
Thanks!

PR #725, opened 2017-10-08
Correct typo in cross-ref. to Passthroughs section

andrewcarver:
At the end of this section on Quotation Marks and Apostrophes (22.2 on the website), there is a link, broken, to the Passthroughs section (43 on the website). It's broken only because it has '#passthru', while the ID of sec. 43 is '#passthroughs'. A quick realignment of the spelling, and  we're "in business" ;)


mojavelinux:
Thanks @andrewcarver!

PR #726, opened 2017-10-08
Correct typo in cross-ref. to Passthroughs section

andrewcarver:
Towards the end of this section on Special Characters (40.1 on the website), there is a link, broken, to the Passthroughs section (43 on the website). It's broken only because it has '#passthru', while the ID of sec. 43 is '#passthroughs'. A quick realignment of the spelling, and  we're "in business" ;)


mojavelinux:
Thanks!

I was wondering why these weren't caught when using the `-v` flag. Then I realized you've discovered another defect. If the reference uses the inter-document reference format, and the reference resolves to the current document, the xref isn't validated. I'll fix that so we can can use the `-v` flag to help find these broken references.

PR #727, opened 2017-10-09
Emend "Attribute value" row in sec. 40.6 table

andrewcarver:
The table in view is entitled, "Elements subject to post replacements text substitution". Its section, 40.6, is entitled, "Post Replacements". Please bear with me as I try to explain the problem as quickly as I can... I'm really just trying to check my understanding of how AsciiDoctor works (which, when it seems not to comport with something in the manual, concerns me!!).

So yes--I must admit, I'm a bit unsure of the exact semantics of this "Attribute value" row, in such tables (in Secs. 40.1-40.6). However, logic seems to constrain a reasonable interpreter thus: 

This row's presence here would make no sense, really, if the row intended to refer either to
1) "Named attributes" (which, according to Sec. 13.5.2, have their own substitution rules, dependent on use of single or double quotes); or to
2) Substitutions within the value that an attribute-reference has already expanded _into_ (after all, whether _that_ substitution would take place depends *not* on whether the value came from an attribute-reference expansion, but rather on what the order of substitutions might be, *post-attribute-ref-expansion*, at the point the attribute-reference occurs).

So, I'm inferring from all this that the row intends, rather, to refer to expansions, within an attribute's value, _at the place where the attribute was defined_. (*Please correct me*, if I'm wrong!)

Add that "fact" to one other I seem to "know", and the current population of this row becomes problematic (to me): This second "fact" is that my tests seem to show, concerning "replacement", "post replacement", and "quote" substitutions, that any of them _can_ happen within the attribute's value, _at the point where an attribute-ref has (already) expanded into that value_; *but*, such expansions can never happen within the attribute value _at the point where the attribute is being defined_. (It seems irrelevant to this, whether the attribute is defined in the doc header or defined later.)

I did all my tests on these expansions with the AsciiDoctor version (apparently 1.5.5-4) that is in the v.2.10.2 "asciidoc-preview" package for Atom. (I gathered this was its AsiiDoctor version, from scanning the package's CHANGELOG in Atom.)

OK--if my interpretation of the semantics were sound, and unless my tests (repeated several times) are tricking me, it would seem to follow that the row under discussion should have a cross (for NO), rather than a check-mark, in column 2. This conclusion seems corroborated by the fact that the corresponding rows in the corresponding tables about Replacements expansions (sec. 40.4) and Quotes (sec. 40.2), seemingly subject to the same interpretation-logic and giving me the same test-results, have crosses in that column.

Anybody follow, and agree with, my logic??
 
By the way, macro-substitution (sec. 40.5) is a horse of a rather different color: I've not been able to get *any* macro-substitution to work in an attribute value, whether at the point of definition or at the point of attribute-ref. However, all the more does it seem to follow, that the corresponding row in sec. 40.5's corresponding table should have a cross in column 2--but, it too has a check-mark. This may indicate that I'm totally off-track in my reasoning; or it may be that that table is actually subject to the same (or a stronger) argument and needs similarly emending. 
I guess that, strictly speaking, that one would be a discussion for a different PR ; but does anyone wish to comment on it in this PR?
Oh, and one other discussion for a different PR ;) Can anybody think of a way to (re)name these table rows, that would clarify better their exact meaning?

### DIFF
--- a/docs/_includes/subs-post.adoc
+++ b/docs/_includes/subs-post.adoc
@@ -11,7 +11,7 @@ The line break character, `{plus}`, is replaced when the `post_replacements` pro
 |===
 |Element | `post_replacements` substitution
 
-|Attribute value |{y}
+|Attribute value |{n}
 
 |Comment |{n}
 


andrewcarver:
If we discover PR #729 to be soundly based, it would add strong corroboration to this one.

mojavelinux:
Thanks for your detailed analysis.

In general, I find these tables to be confusing. I think they are trying to explain how substitutions work by itemizing all applications rather than explaining the logic behind how they are applied.

As I mentioned in the other thread, the trick to seeing what is stored in an attribute value (and hence, what got substituted) is to use a pass block as follows:

```
:name: foo + \
bar

[subs=attributes]
++++
{name}
++++
```

This outputs:

```
foo +
bar
```

So clearly, the post_replacements substitution was not applied to the attribute value itself.

However, if you were to use it in a paragraph context, then the substitution would be applied, but only because it gets applied in a paragraph after the attributes substitution.

```
{name}
```

results in

```
<div class="paragraph">
<p>foo<br>
bar</p>
</div>
```

That should help what's going on.

mojavelinux:
>  I'm a bit unsure of the exact semantics of this "Attribute value" row, in such tables (in Secs. 40.1-40.6). 

My understanding is that this row is showing which substitutions get applied to the value of an (in-document) attribute entry. We know from that other thread it's the header substitution, which is specialcharacters+attributes.

mojavelinux:
> By the way, macro-substitution (sec. 40.5) is a horse of a rather different color: I've not been able to get any macro-substitution to work in an attribute value,

This is an error in the table. Thanks for pointing that out. The only macro that can be used in the value of an attribute entry is the inline pass macro, and it must entirely enclose the value. In other words:

```
:name: pass:r[foo--bar]

[subs=attributes]
++++
{name}
++++
```

which outputs:

```
foo&#8212;&#8203;bar
```

However, the inline pass macro is not used in this case:

```
:name: pass:r[foo--bar] baz
```

andrewcarver:
Thanks for your excellent feedback, Dan.

Given that probably the main ambiguity in the row's semantics is between substitution after the attribute-ref. has been expanded, vs. before then (at the point of definition), one thing I've thought of as a clarification of the semantics is to rename it, 
from "Attribute value" to "Attribute Entry value", or maybe "Attribute Value entry", or "Attribute Value Definition". Do any of those strike you as helpful? 
(I'm happy to submit whatever PRs would make any changes desired.)

mojavelinux:
"Attribute Entry Value" would be perfect. Thank you for offering to submit a patch. I really appreciate your help.

PR #730, opened 2017-10-12
In 40.2-6, rename Row 1 of "Elements subject to"

andrewcarver:
Rename the first row, column 1, of each of these tables, from "Attribute value" to "Attribute Entry Value".

(Cf. #727 re the row's semantics.)


mojavelinux:
Thanks @andrewcarver!

PR #732, opened 2017-10-14
Update description of "Header" subs-group per #730

andrewcarver:
One paragraph in Sec. 40 describes the "Header" substitution group. PR #730 has now made important clarifications about that substitution group, in Sec. 13.4.1. This PR applies a minimal update to the description of it in Sec. 40, using some of the new language of Sec. 13.4.1. 

The sentence I've added here about inline pass macros should especially be scrutinized (it might need massaging ;)


mojavelinux:
LGTM!

PR #733, opened 2017-10-17
Explain how pass-macro "preprocesses" attr values

andrewcarver:
With the aid of Dan's explanation at PR #731, I took a wild stab at broadening Sec. 13.4.1's explanation of how the inline pass macro affects the substitutions happening to an attribute-entry value. 
Emendations (or rejections) expected--and more than welcome!

### DIFF
--- a/docs/_includes/attr-doc.adoc
+++ b/docs/_includes/attr-doc.adoc
@@ -95,10 +95,16 @@ The header substitution group is applied to metadata lines (author and revision 
 This group is also applied to the values of attribute entries, regardless of whether those entries are defined in the header or elsewhere in the document.
 
 If you require different substitutions to be applied to an attribute's value, you can use the <<user-manual#pass-macros,inline pass macro>>.
-In order for the inline macro to be processed, it must completely enclose the attribute value.
-It's used to apply alternate substitutions to the enclosed text at the time the attribute is defined.
+Basically, it works on the attribute's value like a sort of preprocessor. 
 
-The inline pass macro accepts a list of substitutions in the `target` slot.
+It's used to apply alternate substitutions to the enclosed text at the time the attribute is defined.
+The inline macro accepts a list of substitutions in the `target` slot.
+In order for the inline pass macro to be processed, it must completely enclose the attribute value.
+
+An attribute value is checked first for an inline pass macro around it. 
+If the macro is absent, the value is processed with the header substitution group.
+
+If the macro is present, it will control what substitutions are used. 
 In the next example, we'll apply the <<user-manual#quotes,quotes substitution>> to an attribute entry's value.
 
 ----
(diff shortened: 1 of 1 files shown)


docs/_includes/attr-doc.adoc
```diff
@@ -94,11 +94,17 @@ The header substitution group replaces <<user-manual#special-characters,special
 The header substitution group is applied to metadata lines (author and revision information) in the document header.
 This group is also applied to the values of attribute entries, regardless of whether those entries are defined in the header or elsewhere in the document.
 
-If you require different substitutions to be applied to an attribute's value, you can use the <<user-manual#pass-macros,inline pass macro>>.
-In order for the inline macro to be processed, it must completely enclose the attribute value.
-It's used to apply alternate substitutions to the enclosed text at the time the attribute is defined.
+If you require different substitutions to be applied to an attribute's value, there are two approaches from which to choose.
```
mojavelinux: I think saying "two approaches" here is confusing. In fact, I don't think there are two approaches. This is the sole way to change the substitutions that are applied.

andrewcarver: Well, what I was trying to refer to there (though apparently unclearly) were the following:

1) changing the substitutions at the point of attribute-**entry** (via a surrounding inline pass macro)
2) changing the substitutions at the point of attribute-**reference** (via changing the order of subs)
The latter one is mentioned in the text, of course, just past the point where I've edited. 

I take it that I was not clear that this second one was the other "approach" I meant. I'm open to any and every suggestion of ways I could have said it clearer. 
(I don't know if making it sound like these are exclusive alternatives (when in fact you COULD apply both, though it would be odd to WANT to do that) is misleading or not...).

andrewcarver: But I'm starting to see your point now: this whole subsection is entitled, "Substitutions in an attribute entry", so we don't really need to slap all that verbiage in there just to clarify that.

I guess that why I felt this PR would be helpful, was that it clarified for me (based on your explanation) that the pass macro was, here, basically a way to override (via "pre-processing", using your metaphor) the normal set of substitutions that would apply. 

Let me try to clean up that first paragraph I edited -- get it back to what it was, basically, before my last commit. But also, I will suggest one additional word (only) for clarification, in that part I over-edited:

"If you require different substitutions to be applied to an attribute **entry**’s value, ..."

mojavelinux:
Thanks!

mojavelinux:
As a word of advice for future pull requests, you don't to put the full rational for what you changed in the git commit. You can save that for the pull request description. In other words, the git commit message should just describe what was changed. It does not have to defend the change. That's what the PR discussion thread is for.

andrewcarver:
Great, thanks for the tip! And I appreciate, more than I can say, your terrific work with AsciiDoc!

andrewcarver:
WHOA!! Fantastic! I just saw the revision you put up -- MUCH CLEARER discussion! THANKS!!!

mojavelinux:
Thanks Andrew!

mojavelinux:
After editing it, I see why now you had started off saying there are two ways to control the substitution. In the rewrite, I think it's now more clear that there's one way to control the substitutions at assignment and one way to control how it relates to other substitutions when expanded. But I'm open to addition edits if you feel it necessary. In fact, never hesitate to continue making edits. As a writer, I know that editing is never complete. One round of editing just paves the way for future rounds.

PR #734, opened 2017-10-17
Fix table `cols` quickref

pingtimeout:
The quick reference description of the `cols` attribute mentioned  _4 columns_.  But the example table contains only 2 columns and has `cols="2*"`.  Is it a typo?


mojavelinux:
Thanks for pointing this out! I've fixed the problem and improved the description in general.

pingtimeout:
You are welcome, thank you for your work

PR #740, opened 2017-10-28
Replace vbar attribute-ref in col. 2, table in A.3

andrewcarver:
In the table in Appendix A.3, cols. 1 and 2 are styled 'literal'. But a `{vbar}` reference was inadvertently put (not only in col. 3, but also) in col. 2, on the 'vbar' row.


mojavelinux:
It used to be the case that attributes were resolved in literal cells. That was fixed in a recent version. Good catch.

mojavelinux:
Thanks!

PR #741, opened 2017-10-28
Add another case requiring unconstrained quotes

andrewcarver:
Where the User Manual gives us the conditions under which we need to use unconstrained (rather than constrained)  quotes (Sec. 12.3), it lists a couple of marks which, only when they appear directly before the opening quote, require it:

- Is there a colon or semi-colon directly before the starting formatting mark?

But in my experimentation, it seems that a closing curly bracket causes the same necessity. 

This is the case, for example, when you have an attribute reference immediately before the starting formatting mark and you are using quote-substitution before attribute-ref-substitution--as in the normal substitution ordering.

Am I by any chance right about this?? :D


mojavelinux:
That's correct. This character was not being cited.

mojavelinux:
Thanks!

PR #742, opened 2017-10-31
Added plus-based monospace example

bitwiseman:
Asciidoc user guide (http://asciidoc.org/userguide.html#_text_formatting ) shows monospace formatting can be specified using backtick or plus.

### DIFF
--- a/docs/_includes/ex-text.adoc
+++ b/docs/_includes/ex-text.adoc
@@ -89,7 +89,7 @@ _italic phrase_ & __char__acter__s__
 ``++char{plus}++``acter``++s++``
 
 (compat-mode only, deprecated) +monospace phrase {plus} link:.[this page]+ &
-++char{plus}++act{dash}er++s++ 
+++char{plus}++act{dash}er++s++
 
 `*monospace bold phrase*` & ``**char**``acter``**s**``
 
--- a/docs/_includes/ex-text.adoc
+++ b/docs/_includes/ex-text.adoc
@@ -146,7 +146,8 @@ The vial's label read: E=mc^2^; `the scent of science`; `_smell like a genius_`.
 // end::mono[]
 
 // tag::b-mono-code[]
-Reference code like `types` or `methods` inline.
+Reference a `Class` or an `instance{plus}member` inline.
+Also reference ``Class``es or ``instance{plus}member``s.
 // end::b-mono-code[]
 
 // tag::highlight[]
--- a/docs/_includes/ex-text.adoc
+++ b/docs/_includes/ex-text.adoc
@@ -85,8 +85,8 @@ _italic phrase_ & __char__acter__s__
 `monospace phrase {plus} link:.[this page]` &
 ``char{plus}``acter``s``
 
-+monospace phrase {plus} link:.[this page]+ &
-++char{plus}++acter++s++ (plus)
+(compat-mode only, deprecated) +monospace phrase {plus} link:.[this page]+ &
+++char{plus}++act{dash}er++s++ 
 
 `*monospace bold phrase*` & ``**char**``acter``**s**``
 
--- a/docs/_includes/ex-text.adoc
+++ b/docs/_includes/ex-text.adoc
@@ -82,7 +82,11 @@ _italic phrase_ & __char__acter__s__
 
 *_bold italic phrase_* & **__char__**acter**__s__**
 
-`monospace phrase` & ``char``acter``s``
+`monospace phrase {plus} link:.[this page]` &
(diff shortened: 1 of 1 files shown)


mojavelinux:
The plus is the legacy syntax. In Asciidoctor, it means inline passthrough unless you have compat mode turned on. Therefore, the only place we document it is on the migration page.

See http://asciidoctor.org/docs/migration/

bitwiseman:
How does one know what is legacy and what isn't?
Looking at the migration document theres also a "\`+" syntax that is not in the example text.  

Why would asciidoc.org be wrong?

mojavelinux:
> Why would asciidoc.org be wrong?

asciidoctor.org is the canonical information about AsciiDoc, as the Asciidoctor project now manages the definition of the language. Our goal is to overhaul asciidoc.org to be information about just the language itself, but I don't yet the :dollar: for that initiative atm.

Unless you intentionally enable compat mode, you should assume AsciiDoc is being parsed in modern mode. The compat mode is only there to preserve the integrity of old documents.

mojavelinux:
The Jenkins website is running in compat-mode. It really should not be. No new documents should be using compat mode at this point. All major integrations are using the modern syntax now (including GitHub).

https://github.com/jenkins-infra/jenkins.io/blob/master/content/_config/site.yml#L19

bitwiseman:
@mojavelinux - Thank you so much.  I'll update this PR shortly with "\`+" syntax.

jglick:
IIUC `++` is critical for some purposes. At least, I was not able to find any other way to properly render

```asciidoc
Do not pass random ++Object++s to methods taking ++String++s!
```

as

> Do not pass random `Object`s to methods taking `String`s!

mojavelinux:
With the modern syntax, it's:

```
Do not pass random ``Object``s to methods taking ``String``s!
```

In short:

* ` means monospaced
* + means passthrough

(you can combine them, with the passthrough on the inside).

jglick:
OK. Good to know.

bitwiseman:
@mojavelinux - updated with passthrough example and compat-mode example.

bitwiseman:
@mojavelinux - Is this okay to merge?  No hurry, I'm wondering about process.

bitwiseman:
@mojavelinux  Ping, can we merge this?

mojavelinux:
I agree this should be added somewhere in the user manual, but I don't want to add it to the quick reference guide. The reason is, the quick reference guide should only show the modern syntax and shouldn't get too far into the details of edge cases. Therefore, I'll find the right place for it in the user manual, then push an update. If we agree, then we'll proceed with merging.

Just to be clear, the hold up isn't about the change itself, but rather about the decision as to where the change belongs.

bitwiseman:
@mojavelinux 
You're in charge and I'll go with what you decide here, but let me offer a counter argument. 

Two of the three lines aren't compat-mode and also are not what I would call edge cases when documenting APIs and talking about instances of `someClassName`.  "When you have three 'someClassNames`s, this is what happens."  Being clear about these will really help.

I understand your point about the compat mode, however the lack of it being mentioned anywhere in the quick reference, resulted in me spending quite a bit of time banging my head on my desk.  One line out of the whole quick reference isn't really that much to ask is it?

mojavelinux:
I get what you're saying. I'll find a compromise that takes that into account. Stay tuned.

bitwiseman:
@mojavelinux 
Another point here, for this section we're using the same include for both the quick reference and the writers guide.   So, if we want to include this in the writers guide, we end up including it in the quick ref as well. 

I'm not trying to put the work on you.  I'd rather present you an improvement that works. 

There's a lengthy discussion of legacy monospace in https://raw.githubusercontent.com/asciidoctor/asciidoctor.org/master/docs/_includes/text-mono.adoc but is it all commented out.

Another place this could go is https://github.com/asciidoctor/asciidoctor.org/blob/master/docs/_includes/ex-pass.adoc

mojavelinux:
I believe I have addressed the concern with the change made in ed5d6c296e782bbb8c9b3740fb81f367e8d5541d.

This is really tricky because the syntax quick reference needs to be coherent and easy to understand, but at the same time touch on all the key scenarios.

I drew ideas from this patch, but ultimately decided to simplify it. I believe accepting at as is would have made the quick reference less approachable.

Where we could still improve is some more real-world scenarios, and perhaps reordering some of the examples. If you can think of a way to keep the text flowing nicely, but introduce cases that are not covered, I welcome the suggestions. I admit this is not easy, but a good example really pays off.

mojavelinux:
And thank you.

PR #743, opened 2017-11-12
update page to match plugin's README

hboutemy:
particularly https://github.com/asciidoctor/asciidoctor-maven-plugin/pull/322


mojavelinux:
Thanks! I'm going to merge this, but what we really need to do is include the file directly from the Maven repository so we don't have to maintain two versions of it. I'll follow-up by making that change.

PR #748, opened 2017-11-22
Emend :stylesdir: to navigate via parent directory

andrewcarver:
Correct me if I'm wrong, but in this final Sec.-76 code-example, the first directory-move must be to the directory (/mydocuments) which is the parent of the directory (/mynesteddocuments) that contains the current file (c.adoc), since said parent is the directory that contains the directory you're trying to reference (/mystylesheets). So, the single-dot should be a double-dot, in the filepath.


mojavelinux:
Right you are!

mojavelinux:
Thanks!

PR #749, opened 2017-11-28
Add note for unsupported automatic section title resolution

gaganis:
It feels natural to users that section titles for inter-doc references
should be resolved automatically as it also happens for internal cross
references.

This limitation is clearly documented in the User Manual but I believe a
small note is warranted also in the quick reference so people who are
looking at the reference and still try to make this work.

Relates to: asciidoctor/asciidoctor#2500


jirutka:
BTW I wrote an extension that overcomes this limitation, [asciidoctor-interdoc-reftext](https://github.com/jirutka/asciidoctor-interdoc-reftext).

PR #757, opened 2017-12-28
Use the latest version of Asciidoctor in the examples

Mogztter:
We could define an attribute but I didn't know where to define it 😉


graphitefriction:
> We could define an attribute but I didn't know where to define it :wink:

I just revisited the ginormous user manual header :scream_cat: because I had to update something else. OMG, who created this monster ... oh, that was me :flushed: Yes, in the future this would be an awesome thing to have an attribute for (we may even be able to figure out how to automate it), but for now, we'll just merge this in as it is. Thanks!

PR #762, opened 2018-01-15
add table example with colspan and rowspan for the same cell

jmini:


jmini:
I found an example of a cell having `colspan` and `rowspan` set in this blog post:
http://mrhaki.blogspot.fr/2014/12/awesome-asciidoctor-span-cell-over-rows.html

In my opinion such an example should also be present in the user-manual.

mojavelinux:
Thanks @jmini!

PR #777, opened 2018-03-14
Add replacements subs for attributes tables

oncletom:
This way, `wj^[5]^` can be displayed as a footnote link.
At the moment, it is displayed as plain `wj^[5]^`.

### DIFF
--- a/docs/_includes/attrs-charref.adoc
+++ b/docs/_includes/attrs-charref.adoc
@@ -4,7 +4,7 @@
 [#charref-attributes-table]
 // tag::table[]
 .Predefined attributes for character replacements ^[1][2][3]^
-[width="75%", cols="^4l,^3l,^3",subs="replacements"]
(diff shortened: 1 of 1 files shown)


oncletom:
I'm unsure why it does not work though :-(

das-g:
Maybe #794 is an alternative?

mojavelinux:
> I'm unsure why it does not work though :-(

It doesn't work because the literal style was being applied to the cells and substitutions are not applied to literal table cells. @das-g gave me the idea of using a different style. In the end, I choose to apply the monospaced style in the first column and selectively the emphasis style in the second.

The problem is now fixed by a88280f83a3e517a45a2b158d2388c15a63639f9, though it still makes me wonder whether we need more control over substitutions on table cells.

mojavelinux:
I forgot to close this issue as it is now resolved. Closing it now.

PR #778, opened 2018-03-28
Add note about nesting blocks

janicesignalfx:


mojavelinux:
Thank you for submitting this update. This is an important point to emphasize.

There's just one minor change I'd like to request. To make the nested block easier to identify on visual inspection, I prefer to move in increments of two symbols. In other words, instead of suggesting adding one extra symbol, can you please suggest adding two instead? Thanks!

PR #779, opened 2018-03-31
Fix allowed value for :copycss:

rkratky:
The **Allowed values** field for the `copycss` attribute erroneously says *empty*. Instead, it should say that the value needs to specify the location of the stylesheet (provided we don't want the `stylesheet` attribute to resolve relative to the base directory. This PR fixes that.

See also https://github.com/asciidoctor/asciidoctor/issues/300#issuecomment-26478893.

### DIFF
--- a/docs/_includes/attrs-builtin.adoc
+++ b/docs/_includes/attrs-builtin.adoc
@@ -878,7 +878,7 @@ _Introduced in 1.5.6._
 |copycss
 |If set, copy the CSS files to the output.
 |_empty_ (File is copied if `linkcss` is set).
-|_empty_
+|Location of the stylesheet
 |{y}
 |<<applying-a-theme>>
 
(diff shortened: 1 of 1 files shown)


mojavelinux:
Technically, empty is a valid value if you are using the default (built-in) stylesheet. But you are correct that if you are using a custom stylesheet, it needs to be the stylesheet you want to copy. So it's really both.

rkratky:
Right, so perhaps it should say 'empty or ...'? Because saying just 'empty' is confusing -- it sounds as if nothing but empty is allowed.

rkratky:
So, specifically, "Empty or the location of the custom stylesheet (if used)".

mojavelinux:
You got it!

mojavelinux:
Thanks @rkratky!

PR #784, opened 2018-04-27
Update reference to http to be https

NottyCode:
When describing the hide-uri-scheme attribute the documentation uses an https url, but then says note the absence of http in the url. Given it is https I expect this was an oversight.

I'm sorry if there are contribution guidelines I'm supposed to follow I couldn't find any.


mojavelinux:
Right you are! Thanks @NottyCode.

PR #785, opened 2018-04-29
Emend demo code for registering treeprocessor

andrewcarver:
I don't think having a |document| parameter is correct here, for the code-block passed to Asciidoctor::Extensions.register ...?? (It didn't work when I tried it, anyway--but I'm not sure why! :)

### DIFF
--- a/docs/_includes/exten-tree.adoc
+++ b/docs/_includes/exten-tree.adoc
@@ -61,7 +61,7 @@ end
 .Usage
 
 ```ruby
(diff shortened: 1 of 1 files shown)


mojavelinux:
You are correct. This is a remnant of an earlier design. The document is accessible via the registry, which is the optional first argument of the block.

```ruby
Asciidoctor::Extensions.register do |registry|
  puts registry.document
end
```

mojavelinux:
Thanks!

PR #787, opened 2018-05-03
remind 'em that doctype=book implies chap.=level 1

andrewcarver:
Probably there's a better wording; someone please supply it.


mojavelinux:
I like the point you're making here. I rephrased it slightly.

mojavelinux:
Thanks!

PR #788, opened 2018-05-07
transpose sections 72 and 73

andrewcarver:
Put the material in the (logical) order it appeared in the README.adoc in the Stylesheet Factory GitHub page (https://github.com/asciidoctor/asciidoctor-stylesheet-factory). After all, you can't know even what folders the current section 72 is talking about, till you've started Sec. 73. It wasn't until I read the GitHub page that I could see that the User Manual had somehow gotten these sections out of the didactically logical order that appears on GitHub. There might be a few tweaks that could help, in addition to this transposition of the sections; but just the transposition, with nothing else done, would make the material vastly more intelligible--IMHO.

### DIFF
--- a/docs/user-manual.adoc
+++ b/docs/user-manual.adoc
@@ -1194,16 +1194,6 @@ The Asciidoctor project includes alternative stylesheet themes from <<stylesheet
 You can also create your own <<custom-themes,themes>> and <<custom-backends,backends>>.
 --
 
-== Custom Themes
-
-=== Creating a Theme
-
-include::{includedir}/create-theme.adoc[]
-
-=== Applying a Theme
-
-include::{includedir}/apply-theme.adoc[]
-
 == Stylesheet Factory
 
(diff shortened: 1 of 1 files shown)


mojavelinux:
I agree the current order is confusing. However, I think we should make a more specific change. Let's move "Applying a Theme" to the first chapter under the "Customizing Your Output" part (moving it up a level). The reason is, this section actually has nothing to do with the stylesheet factory. It does assume you have a stylesheet to apply, but that's the intended audience.

Then, we move "Creating a Theme" under the "Stylesheet Factory" chapter, before "Applying a Stylesheet". That section is definitely part of the stylesheet factory instructions.

I recognize that some of the titles here need work, but I'm ignoring that for right now.

andrewcarver:
Yes--perfect! I've now tried to capture that...

mojavelinux:
Thanks @andrewcarver!

PR #791, opened 2018-05-30
Fix attribute reference in preface note

slyfoxza:
When the IRC link was changed to Gitter, the `uri-irc` reference was left in the note, resulting in the markup rendering verbatim.


mojavelinux:
Oops! Thanks for correcting this, @slyfoxza!

PR #794, opened 2018-06-06
fix table end note markup for ZSWP and WJ rows

das-g:
fix table end note markup for `ZSWP` and `WJ` rows

(Though I wonder whether https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#section-footnotes should be used here instead of manually applying superscript.)


mojavelinux:
Thanks for pointing this out!

I decided to fix it by making the first column monospaced. I also fixed formatting in the second column as well.

In this case, I choose to use explicit markup rather than footnotes to keep the notes close to the table and self-contained. In the future, we might be able to control where footnotes get emitted, which will allow us to switch back to using footnotes and accomplish the same result. See https://github.com/asciidoctor/asciidoctor/issues/1705

mojavelinux:
Resolved by a88280f83a3e517a45a2b158d2388c15a63639f9

PR #795, opened 2018-06-18
fix typo in docs/_includes/ex-dlist.adoc

Siltaar:
A `second term` has been mistakenly spelled `section term`.

(Else the french translation on which I'm working on is 50% complete, regarding the asciidoc-syntax-quick-reference… it's a slow work, but I'm not alone to work on)


mojavelinux:
Thanks for the fix! And I'm excited to hear the translation is coming along.

Issue #797, opened 2018-08-04
The Asciidoctor Eclipse plugin with live preview


charleech:
Hi,

I'm not sure if you're aware of  Eclipse plugin named [Asciidoctor Editor](https://marketplace.eclipse.org/content/asciidoctor-editor) or not. Their GitHub is [de-jcup/eclipse-asciidoctor-editor](https://github.com/de-jcup/eclipse-asciidoctor-editor).

I've downloaded and tested, it is very impressive as same as[ AsciidocFX](https://asciidocfx.com/).

Please correct me if I'm wrong. It would be great to add that plugin to the [Editing AsciiDoc with Live Preview](https://asciidoctor.org/docs/editing-asciidoc-with-live-preview/)

mojavelinux:
I agree. If you'd like to see it mentioned on that page, feel free to propose a change. Thanks!

charleech:
Please review the #798.

mojavelinux:
This is now live.

PR #798, opened 2018-08-04
Add Eclipse Asciidoctor Editor plugin.

charleech:
Regarding to #797 

Add Eclipse Asciidoctor Editor plugin.


mojavelinux:
This has been superseded by #807. Since that entry is more complete, I'm going to go with that one.

PR #802, opened 2018-08-28
Fix trivial syntax error in User Manual

owenh000:
Add missing space in User Manual `html-styles.adoc`, which prevents
correct page generation at
<http://asciidoctor.org/docs/user-manual/#styling-the-html-with-css>.


mojavelinux:
Great catch, @owenh000. Thanks!

PR #809, opened 2018-09-29
Add htmlbook and link the asciidoctor-backends repo to template examples

danyill:
These examples would have helped me a few weeks ago, I thought we should add them to the docs.
The test suite has almost no examples but asciidoctor-backends is like gold for someone wanting to use templates.


mojavelinux:
Sounds good to me. Thanks!

PR #810, opened 2018-09-30
Image macro documentation for svg images added

### LINKED ISSUE #476: Document SVG embedding options for HTML5 converter
Document the SVG embedding options for the HTML5 converter.
- _no option_ (default)
- interactive
- inline

Note that when using inline, there are several requirements that must be met in order for the SVG to work properly when included inside an HTML5 document.
- the primary SVG elements (e.g., `<svg>`) cannot have a namespace
- ideally the SVG should be optimized to minimize the content being put into the HTML document

When using inline or interactive, the following requirements must be satisified:
- the `viewBox` attribute must be defined on the `<svg>` element for scaling to work properly

When using the inline option, if you specify a width or height in the image macro in AsciiDoc, the width and height attributes on the `<svg>` element will be removed.

For more details, see https://github.com/asciidoctor/asciidoctor/issues/1301.

danyill:
Close #476


mojavelinux:
Thanks!!

Issue #745, opened 2017-11-17
Clarify which substitutions take place in attribute list


andrewcarver:
There needs to be a better attempt at explaining, in Secs. 10.5.1 and 10.5.2, exactly what substitutions take place within Attribute Lists' values. If you're looking for an example of what I mean, please see the crystal-clear explanation of this topic in the AsciiDoc User Guide, [Sec. 14.1](http://asciidoc.org/asciidoc.css-embedded.html#_attribute_value_substitution).

It goes without saying, that a user of AsciiDoctor should not have to go to the AsciiDoc User Guide to get an explanation of how AsciiDoctor works--not least because it's not about AsciiDoctor. So, I won't stoop to mentioning it. :D

mojavelinux:
The main issue at hand is that there is no section dedicated to covering block attributes. Block attributes are treated sort of like an extension to document attributes, but they're profoundly different. How they are defined is different and how they get used is different. The only real relationship between them is that certain block attributes inherit a default value from a document attribute, but that's not very common and may be deprecated in all be special cases.

andrewcarver:
Ah-- I believe you've smacked the nail on the head. (Sec. 10.5 is, I take it, the section you feel is "sort of floating there"... )

I can suggest here a wee initial step: in AsciiDoc an "AttributeList" is a block element; yet, in our list of such [https://asciidoctor.org/docs/user-manual/#elements](url), we omitted it. It should be there, no?

andrewcarver:
In my all-too-deep ignorance on this issue, which I'm trying to mitigate, please allow me float yet another take on it--which I hope you'll shoot down just as quickly, if it's likewise faulty:

Aren't you implying, that among the six types of "attributes" listed at the start of [section 10](https://asciidoctor.org/docs/user-manual/#attributes), those called "element attributes" (for which, I take it, you prefer the name "block attributes") are so fundamentally different, as to be really a different category from the others?

andrewcarver:
Why else would we both feel that Sec. 10.5 is out of its natural element there?

andrewcarver:
Thus, if I'm understanding, "document attribute" and "block attribute" are the two basic kinds of attribute. And if so, it seems logical to me that we should have a top-level section for each of these.

As the [User Manual's glossary](https://asciidoctor.org/docs/user-manual/#glossary) defines them:

> **block attribute**
[...]

PR #812, opened 2018-10-02
Add 'Attribute List' to the list of block elements

andrewcarver:
Attribute List is, I suspect, an item that should be in our User Manual's list of block elements. It is mentioned in the AsciiDoc User Guide's [discussion of such](http://asciidoc.org/asciidoc.css-embedded.html#_block_elements), and its importance as a block element seems underscored by the (ongoing) discussion at #745.

### DIFF
--- a/docs/_includes/element.adoc
+++ b/docs/_includes/element.adoc
@@ -29,6 +29,7 @@ A document can include the following block elements:
 * Block Title
 * Block Macro
 * Block
+* Attribute List
 * Paragraph
 * Delimited Block
 * Table


mojavelinux:
Actually, this is incorrect. The attribute list is most definitely not a block element. I always felt that the AsciiDoc User Manual got this wrong (or perhaps Stuart was defining a block element differently).

An attribute list is a member of every block element. The entries are defined using one or more block attribute lines.

mojavelinux:
I'm not even really sure listing elements this way is even helpful for users. This is really something that belongs in an AsciiDoc spec, not a user manual, IMO.

andrewcarver:
Ah. I stand corrected; thanks for the clarification.

mojavelinux:
Sorry for not providing the right information before. This is actually a pretty tricky topic to explain, so it might take some trial and error to get it right. And the user manual in its current state has contributed to the confusion.

andrewcarver:
No need to apologize. Your "document attribute" vs. "node attribute" makes sense, so that's a solid foundation.

PR #814, opened 2018-10-16
Strengthen to +++ the + around mention of pass:[]

andrewcarver:
In Sec. 10.4.3 of the User Manual, end of first sentence, someone attempted an inline mention of `pass:[]` by embedding it in a single-plus passthru macro: `+pass:[]+`. But for whatever reason, single-plus doesn't suffice with this content: it produces only an empty `<code></code>` element in the output.

Using the triple-plus passthru does work, however: `+++pass:[]+++`.


andrewcarver:
Thanks Dan, for the fix and for the explanation--patient, helpful, useful.

PR #823, opened 2018-12-20
List extension for rendering tabbed code blocks

bmuschko:
The extension is well documented on GitHub. Based on the [discussion on Twitter](https://twitter.com/bmuschko/status/1075398807829983233), other users seem to be interested in the functionality.


mojavelinux:
Thanks!

FYI, I also implemented tabsets in Antora. Here's the approach I took:

https://docs.couchbase.com/home/contribute/tabs.html

Here's the code: https://gitlab.com/antora/antora-asciidoctor-extensions/blob/master/tabs-block/extension.js

I prefer the nested code style instead of role correlation, though the extension could implement both structures.

PR #826, opened 2018-12-26
Update ex-rev.adoc

CodeFerriCode:
Add V in `revnumber` to make it consistent with: (asciidoctor.org/docs/_includes/revision.adoc)

- `In the converted document, notice that the V preceding the revnumber is capitalized in the byline but not when the attribute is referenced in the body of the document.`
- `When explicitly set, any characters preceding the version number are not dropped.`


mojavelinux:
A similar change was made to the new docs site. See https://docs.asciidoctor.org/asciidoc/latest/document/reference-revision-attributes/

PR #827, opened 2018-12-26
Add missing metadata name `citetitle`

CodeFerriCode:
Add the missing metadata name `citetitle` to the listing before the example.


mojavelinux:
Technically, cite title is just one of several block attributes.

We could mention that certain blocks have additional positional attributes. But that would need to come after the initial list show here.

PR #834, opened 2019-01-26
include puravida software extensions

jagedn:
I've included a list with some extensions I developped 

Most of them are in an stable version except *quiz* that it's in an `experimental` state


mojavelinux:
Thanks @jagedn! Great work.

(Sorry for the delay).

PR #841, opened 2019-03-01
Update website ref: Asciidoctor-reveal.js 2.0.0 is released

obilodeau:
Took the liberty to handle the FIXME. The link works so it looks ok.


mojavelinux:
:+1:

PR #859, opened 2019-04-17
Add extension that outputs compound block

ecerulm:


mojavelinux:
Merged as 856bf9a43905951830d40734311ce8226699ed94. Thanks!

I refactor the extension to reuse the built-in behavior of example block (and thus avoid the use of passthrough content). I also added a few features. You might be interested in checking out the result.

PR #860, opened 2019-04-23
Remove DocBook 4.5 references from documentation.

bneumeier:
Since the current version of Asciidoctor does not have a backend for
DocBook 4.5, remove the assertions that it *does* from the documentation.
also add a reference to the community-supported plugin that restores
support for that format.

I think I found all the mentions of DocBook 4.5 as a current backend in
the docs; apologies if I missed any.


mojavelinux:
Thanks @bneumeier! This looks great.

PR #863, opened 2019-05-15
make markdown URI reference example more idiomatic

matklad:
In markdown, URI references are usually put **after** the main content. Let's do this in comparison, to show that in asciidoctor you have to put it before

Am I correct that it's impossible to achieve markdown behavior with asciidoctor? When writing blog posts in markdown, I tend to have a block of links after each paragraph. I wonder what is the suggested approach to this in asciidoctor? Specifying URLs inline is bad, because urls are long and distract from content.

Specifing `:attr: http://long-url` *before* paragraph works, but feels ... backwards? References and footnotes are usually specified after the content.

### DIFF
--- a/docs/asciidoc-vs-markdown.adoc
+++ b/docs/asciidoc-vs-markdown.adoc
@@ -436,9 +436,9 @@ a|
 a|
 [source,markdown]
 ----
-[home]: https://example.org "Home"
-
 Go [home].
+
+[home]: https://example.org "Home"
 ----
 a|
 [source,asciidoc]


mojavelinux:
I think it's very reasonable to add a basic example that shows the difference between Markdown and AsciiDoc links.

> Am I correct that it's impossible to achieve markdown behavior with asciidoctor?

No, that's not correct. You can use a custom inline macro to emulate the behavior of Markdown. In fact, I've been meaning to push this to the extensions lab, so I'll follow up by doing that.

> Specifing `:attr: http://long-url` before paragraph works, but feels ... backwards? References and footnotes are usually specified after the content.

This is just a matter of opinion (and all what you are used to). I find it unnatural to do it the other way around. Having said that, you may be interested in this proposal: https://github.com/asciidoctor/asciidoctor/issues/395

mojavelinux:
The extension I mentioned is now available. See https://github.com/asciidoctor/asciidoctor-extensions-lab/blob/master/lib/markdown-link-inline-macro.rb

mojavelinux:
Here's the Markdown link macro for Asciidoctor: https://github.com/asciidoctor/asciidoctor-extensions-lab/blob/master/lib/markdown-link-inline-macro.rb

That extension doesn't currently parse the references, though it could.

mojavelinux:
This was merge as 90bd60a9aaddb71abab2884545d9bacd73edd3e0. Thanks!

PR #868, opened 2019-07-15
Installing the asciidoctor-firefox-addon.xpi fails

jnweiger:
Firefox 67.0 on Linux Mint says the xpi here is corrupt. As a workaround, I am adding the link to the mozilla.org site, the live preview pluging there works fine for me.

You may want to fix this differently, of course.


mojavelinux:
This page has been replaced in the new docs site and information about how to install the extension is now maintained by the extension project itself. Therefore, this patch is now obsolete.

PR #877, opened 2019-08-05
Add instructions on using an include with content on a list item (clo…

### LINKED ISSUE #876: Document adding content to a list item
As per http://discuss.asciidoctor.org/Includes-in-part-of-a-line-includes-in-numbered-lists-td7039.html#a7040

PR on the way.

danyill:
Closes #876

### DIFF
--- a/docs/_includes/include-list-content.adoc
+++ b/docs/_includes/include-list-content.adoc
@@ -1,0 +1,25 @@
+To use the include directive with content as part of an existing line requires a small trick.
+The `include` directive must be on "it's own" line.
+However the list item must have some content to be recognized as a list item.
+To achieve this we use the `\{blank}` built in attribute (see <<complex-list-content>> for another example of this) to give the list item some content.
+
+So to have an `include` add content to a numbered list item we can use:
+
+[source,asciidoc]
+----
+. {blank}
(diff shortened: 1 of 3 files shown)


mojavelinux:
@danyill I applied some revisions (okay, a lot). Let me know what you think.

It's important to distinguish the primary text of a list item from the complex content. I tried to do that in my edits.

danyill:
LGTM. Thank you for taking the time to review. I felt uncomfortable using a word like a "trick" but couldn't think of a better way of expressing it. I think what you've written is clearer.

Out of curiosity, while are you fully qualifying the xref against the user manual (e.g. `<<user-manual.adoc##dropping-the-principal-text>>`) here?

mojavelinux:
Thanks for reviewing, @danyill!

> Out of curiosity, while are you fully qualifying the xref against the user manual (e.g. <<user-manual.adoc##dropping-the-principal-text>>) here?

Technically, it's not necessary to make the reference work. But we're doing it that way so that it can be more easily migrated to Antora. Once we've made that switch, it will become the page ID you recognized.

mojavelinux:
Btw, I also removed the other explicit IDs for the same reason. Once we're in Antora, everything will get reworked.

PR #881, opened 2019-08-22
Document the opts=optional attribute which can be used for an inclu…

jwillemsen:
…de directive and fixed one typo

### DIFF
--- a/docs/_includes/include-directive.adoc
+++ b/docs/_includes/include-directive.adoc
@@ -38,10 +38,10 @@ The include directive has the following anatomy:
 
 [source,subs=+quotes]
 ----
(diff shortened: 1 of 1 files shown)


mojavelinux:
Awesome! Thanks!

It needs one correction before it can be merged. The optional option is specified using `opts=optional` (not `optional-option`). Could you update?

Also, would you be willing to add a sentence that explains that when the optional option is set, the processor will ignore the directive if the target cannot be found? (meaning no warning is shown).

jwillemsen:
I will make an update tomorrow and test your proposed change, optional-option does seem to work

mojavelinux:
> optional-option does seem to work

While I agree that's true, it's matching the internal name used to track the option, and thus should be avoided.

mojavelinux:
Thanks!

PR #885, opened 2019-09-01
google+ no longer active

ahus1:
This change removes the link from the template. 
I was unable to test this locally due to general problems with rake in my installation. 
I hope Netlify will provide a preview.


mojavelinux:
Yep, preview looks good!

PR #821, opened 2018-12-10
Updated migrating-from-msword.adoc to pandoc 2.0

abelsromero:
Related to https://github.com/asciidoctor/asciidoctor.org/issues/542

This PR:
- Removes the `--normalize` flag, no longer necessary
- Updates tested on software
- Added information about links compatibility
- Added link to @Mogztter GitHub on mention
- Fixed a couple of typos
- Removes unused attribute `uri-ant`

PR #894, opened 2019-10-04
Add hybrid list with inner definition list example

janicemhdc:
This is not obvious so adding an example of how to do this to the list docs.

### DIFF
--- a/docs/_includes/dlist.adoc
+++ b/docs/_includes/dlist.adoc
@@ -80,5 +80,19 @@ Here's how the list is rendered:
 include::ex-dlist.adoc[tag=3-mix]
 ====
 
+In order to use definition lists inside ordered or unordered lists you must use an \{empty} attribute after the ordered or unordered list marker then place the two portions of the definition list on the next line as follows:
+
+[source]
+----
+include::ex-dlist.adoc[tag=inner-dlist]
+----
+
+Here's how the list is rendered:
+
+.A hybrid list with an inner definition list
+====
+include::ex-dlist.adoc[tag=inner-dlist]
+====
+
 You can include more complex content in a list item as well.
 // end::hybrid[]
--- a/docs/_includes/ex-dlist.adoc
+++ b/docs/_includes/ex-dlist.adoc
@@ -93,6 +93,15 @@ Cloud Providers::
     . Rackspace
 // end::3-mix[]
 
+// tag::inner-dlist[]
+* {empty}
+Booleans:: Use true and false, not 1 and 0 or T and F
+* {empty}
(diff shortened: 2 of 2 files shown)


mojavelinux:
Thanks for proposing this, @janicemhdc.

Before I merge, I'm trying to understand what you're after. Do you want each term to be prefixed with a list marker? Is this different than bolding the primary text?

```
* *Booleans*
+
Use true and false, not 1 and 0 or T and F
* *Numbers*
+
Use Arabic numerals without punctuation (other than a decimal point if needed), or abbreviations
* *Enumerated Values*
+
Use one of the allowed values exactly as listed including case.
```

If that's the case, perhaps you are interested in this hybrid list https://github.com/asciidoctor/asciidoctor/issues/821, which I just merged into Asciidoctor PDF https://github.com/asciidoctor/asciidoctor-pdf/issues/1280.

In short, you'd be able to write this list as follows:

```
[unordered.stack]
Booleans:: Use true and false, not 1 and 0 or T and F
Numbers:: Use Arabic numerals without punctuation (other than a decimal point if needed), or abbreviations
Enumerated Values:: Use one of the allowed values exactly as listed including case.
```

If that's what you're looking for, I'd rather implement that then document the workaround. Or we could document that this is a workaround until the feature is implemented.

janicemhdc:
This was in response to https://github.com/asciidoctor/asciidoctor/issues/3445 where I tried and failed to use a dl inside an ordered list. I had looked at the documentation and failed to see any indication of needing this {empty} syntax, hence my flailing around without getting something working. Given that, and that I found the syntax odd, I decided to dash off some quick documentation.

I'd actually prefer not to have the bold text if given a choice. I had used a workaround like the one you presented (without the bold) when the inner DL did not work, but logically it's a bulleted/numbered descriptive list so I'd rather tag it as such.

If I'm understanding #821 correctly it does appear to be what I want. Sorry, I did not come up with that in my search of existing issues.

When is that getting released in an update?

janicemhdc:
Is there an option to not bold the term component? As noted above, I'd prefer that if possible.

mojavelinux:
Do you mean once this is implemented, is there a way to style the term differently? If that's the question, then the answer is yes, the bold style is controlled by the stylesheet / theming system. The bold is not actually part of the semantics of the list.

janicemhdc:
Implementing custom styles is well beyond the time commitment many folks can put in; I've yet to have the resources to do it at any of the jobs where I've used Asciidoctor. 

In most cases I can control whether something is bold or not by using markup or directives; I'm not sure why this would be different. Regardless, thanks for implementing the feature.

PR #895, opened 2019-10-07
Update extensions.adoc

uniqueck:
add a new extension for asciidoctorj

### DIFF
--- a/docs/extensions.adoc
+++ b/docs/extensions.adoc
@@ -49,6 +49,10 @@ Instead, the quality should be based on the extension's own merit.
 |====
 |Name |Runtime |Description
 
+|https://github.com/uniqueck/asciidoctorj-office-extension[Asciidoctor Office]
+|AsciidoctorJ
+|Provides features to convert excel sheets to asciidoc tables and powerpoint slides to images.
+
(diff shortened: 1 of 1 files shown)


docs/extensions.adoc
```diff
@@ -49,6 +49,10 @@ Instead, the quality should be based on the extension's own merit.
 |====
 |Name |Runtime |Description
 
+|https://github.com/uniqueck/asciidoctorj-office-extension[Asciidoctor Office]
+|AsciidoctorJ
+|Provides features to convert excel sheets to asciidoc tables and powerpoint slides to images.
```
mojavelinux: Can you write "asciidoc" as "AsciiDoc", excel as "Microsoft Excel" and powerpoint as "Microsoft PowerPoint"? Thanks!

mojavelinux:
Very cool. I added a request inline to address before merging.

docs/extensions.adoc
uniqueck: Ok is done.
For the missing microsoft part, my idea was to also support Open Document Format, but that is not so easy. So I think for the first time the Microsoft Format is ok.

mojavelinux: I think that's a good goal. The reason I requested adding "Microsoft" is because that's the proper name, not just "Excel" (so it's legally required to write it that way).

PR #903, opened 2019-11-19
Hopefully accurate pointer to Roles and IDs

dblevins:
Here's where I looked for ways to specify the CSS class or id in asciidoctor syntax.  I was falsely confident that if it wasn't mentioned here it likely didn't exist.

In the end I'm using JBake and don't need the stylesheet factory.  I made the note that the stylesheet factory wasn't necessary as I'm the kind of person who would learn it was necessary as this was preceded by instructions on how to set the stylesheet factory. (yes, it's super easy for me to take a wrong turn!! :) )

### DIFF
--- a/docs/produce-custom-themes-using-asciidoctor-stylesheet-factory.adoc
+++ b/docs/produce-custom-themes-using-asciidoctor-stylesheet-factory.adoc
@@ -152,6 +152,13 @@ include::{includedir}/apply-theme.adoc[]
 
 Happy theming!
 
+== Specify CSS class or id
+
+An important feature to learn when customizing a CSS is that the desired CSS `class` or `id` can be specified on Asciidoc elements via 
+a {user-ref}/#role[Role] or {user-ref}/#id[ID].
+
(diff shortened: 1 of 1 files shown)


docs/produce-custom-themes-using-asciidoctor-stylesheet-factory.adoc
```diff
@@ -152,6 +152,13 @@ include::{includedir}/apply-theme.adoc[]
 
 Happy theming!
 
+== Specify CSS class or id
+
+An important feature to learn when customizing a CSS is that the desired CSS `class` or `id` can be specified on Asciidoc elements via
```
mraible:
```suggestion
An important feature to learn when customizing a CSS is that the desired CSS `class` or `id` can be specified on AsciiDoc elements via 
```
:)

dblevins: Thank you!

mojavelinux:
Picking up on your cue, I added this information to the new Asciidoctor docs where I think it fits best (for now). See https://docs.asciidoctor.org/asciidoctor/latest/html-backend/default-stylesheet/#customize-targets

FYI, the stylesheet factory is dead. Long live extending the default stylesheet (or going at it on your own).

PR #904, opened 2019-11-30
updating links in user manual

ahus1:
I found a broken anchor `custom-themes` and ended up fixing some links and upgrading them to https. 

To indicate to other that other chapters references this chapter I've added an explicit reference `[#applying-a-theme]` for it.


mojavelinux:
Thanks!

PR #906, opened 2019-12-04
Register asciidoctor-extension-apidoc

rquinio1A:
I've enriched the idea from https://github.com/asciidoctor/asciidoctor-extensions-lab/blob/master/lib/implicit-apidoc-inline-macro.rb to make it an actually usable extension !
Not sure about the ordering in the table, so I'm adding it last.


rquinio:
Any chance of having the plugin listed in the third-party extensions ? :)
That's probably the only chance for anyone to find it relatively easily, given "apidocs" is not super distinctive name and asciidoctor+javadoc search typically points to asciidoclet...

PR #911, opened 2019-12-22
fix word repetition in callout-xml.adoc

FabienLaunay:
Hi Dan, thanks for merging [PR#910](https://github.com/asciidoctor/asciidoctor.org/pull/910).
To enhance readability of the user guide, I also deleted the repeated article 'the' in the Callouts chapter.

PR #912, opened 2019-12-22
fix word repetition in language-support.adoc

FabienLaunay:
Hi Dan, thanks again for merging [PR#911](https://github.com/asciidoctor/asciidoctor.org/pull/911). Glad to hear you love these :-)
To enhance readability of the user guide, I also replaced the article 'the' by the word 'to' in the Using Asciidoctor with Other Languages chapter.


mojavelinux:
These are so hard to catch without doing a full read, so having you catch and fix them as you come across them is very valuable. Thanks!

PR #913, opened 2019-12-25
Update documentation for callouts in HTML code

luca-heitmann:
I tried using callouts in HTML source code, but the syntax is a little different, as I found out in this issue: asciidoctor/asciidoctor#1608. I thought this could be documented


mojavelinux:
You got it! Thanks!

Issue #914, opened 2020-01-02
Typo in user-manual: "interpets"


mirogta:
... should say "interprets"

mojavelinux:
Thanks!

PR #915, opened 2020-01-02
#914 Fixed typo

mirogta:


mirogta:
Space
Galaxy
Consortium

Unfortunately GitHub doesn't support attaching SVG, which is the format I've set the PlantUML diagrams to export to (so they can be zoomed and text selected/copied), so I've regenerated it in PNG just for this PR. so you can see the result. It doesn't look as nice in PNG as in SVG though.

mojavelinux:
Thanks for the fix!

PR #927, opened 2020-02-19
Asciidoctor reveal.js version bump

obilodeau:
4.0.0 was missing the npm package by mistake so I had to do a 4.0.1 because apparently you can't replace an unpublished npm package even 5 minutes after the initial upload


mojavelinux:
Great work!

PR #929, opened 2020-02-21
Git should ignore files used by rbenv

obfischer:
It would ease the live of people who use rbenv to manage their Ruby installations, if files used by rbenv as `.ruby-version` would be ignored by Git.


mojavelinux:
Seems reasonable to me. Thanks!

PR #935, opened 2020-03-25
Fix markdown syntax for cross referencing

yous:
`link:#_usage[Usage]` is AsciiDoc syntax.


mojavelinux:
Wow, you're totally right. Not sure how we missed that. Thanks!

PR #944, opened 2020-05-19
Document that tags need not be unique 

oddhack:
and that there is defined behavior in this case, per feedback from @mojavelinux on Asciidoctor::Discussion.


mojavelinux:
I have incorporated this change into the new docs. Thanks!

PR #957, opened 2020-06-18
Add reference for using roles inline

ClaytonHughes:
Hi. I always forget the syntax for this when I need it, and thought I'd suggest adding it to the Quick Reference.

Cheers,


mojavelinux:
I've added example to the quick reference in the new docs site.

mojavelinux:
See https://docs.asciidoctor.org/asciidoc/latest/syntax-quick-reference/#ids-roles-and-options

PR #958, opened 2020-06-19
Fix typo SAFE --> UNSAFE

vikram-redhat:
As per my tests, and the rest of the docs, when run from the CLI, AsciiDoctor runs in UNSAFE mode. 

https://asciidoctor.org/docs/user-manual/#set-the-safe-mode-in-the-cli
"When Asciidoctor is invoked via the CLI, the safe mode is set to UNSAFE by default." This is indeed correct as per my tests.


mojavelinux:
That's correct.

You can confirm this using the following command:

```
echo {safe-mode-name} | asciidoctor -o - -d inline -e -
```

PR #966, opened 2020-08-18
Fix typo in unordered list example

petrkajzar:
This is only to fix small typo in Edgar Allan Poe's name in AsciiDoctor unordered list example.


mojavelinux:
Right you are! Thanks!

PR #967, opened 2020-08-19
Fix small typos in the text of Writer's Guide

petrkajzar:
These are only small fixes of minor typos that I've found in the text while studying it. :-)


mojavelinux:
Thanks!

PR #969, opened 2020-09-09
Fix a typo s/Fist/First/

dgoldssfo:


mojavelinux:
Good catch! Wow, that sure was a funny typo.

PR #982, opened 2020-12-28
Fixed Rouge install instructions for Debian-based systems

lgiordani:
I fixed the Rouge installation instructions in the manual as they weren't working on my Ubuntu. I can't check on a pure Debian but I double-checked the name of the package in the following archives

https://packages.debian.org/search?keywords=ruby-rouge
https://packages.ubuntu.com/search?keywords=ruby-rouge&searchon=names&suite=all&section=all

Thanks for Asciidoctor, a great piece of software!

### DIFF
--- a/docs/_includes/src-rouge.adoc
+++ b/docs/_includes/src-rouge.adoc
@@ -13,9 +13,9 @@ Install using `gem` (all systems)::
 +
  $ gem install rouge
 
-Install using `apt-get` (Debian-based systems)::
+Install using `apt` (Debian-based systems)::
 +
- $ sudo apt-get install rouge
+ $ sudo apt install ruby-rouge
 
 Install using `dnf` (Fedora-based systems)::
 +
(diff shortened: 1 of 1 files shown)


mojavelinux:
Thanks! My understanding is that apt is not yet universal (compared to apt-get). Could you clarify which one is the recommended command?

lgiordani:
You are right, I changed that without thinking. It looks like `apt` hasn't replaced `apt-get`, but is intended more as a visually pleasant wrapper (see http://manpages.ubuntu.com/manpages/hirsute/en/man8/apt.8.html). I'm not fussed about that, and since `apt-get` will continue working I'm going to revert that part of the change. Thanks!

mojavelinux:
The old user manual has been replaced and I applied this issue to the new docs. See https://docs.asciidoctor.org/asciidoctor/latest/syntax-highlighting/rouge/#install-rouge

PR #987, opened 2021-01-07
Add missing 'is'

cbachhuber:
Three characters, not my biggest contribution :laughing:


mojavelinux:
Thanks. Except this user manual is no longer being published. It has been replaced by https://docs.asciidoctor.org/asciidoc/latest/directives/include/#include-resolution. And the bug still exists there. Can you click on "Edit this Page" and submit the fix?

cbachhuber:
Done, thank you for the guidance!

PR #994, opened 2021-02-20
List asciidoctor-p2e extension

bosco:
This extension adds a block that you can use to create Pathfinder 2E statblocks. More info about the extension can be found at https://bosco.srht.site/asciidoctor-p2e.html and https://using.tech/posts/p2e/.


mojavelinux:
Cool idea!

PR #998, opened 2021-02-23
Redirect Asciidoctorj docs to docs.aciidoctor.org

abelsromero:
Note 'development' & 'project-layout' are the same un purpose.


abelsromero:
This is not my area, so I just follow the pattern, but I wonder is there's a way to remove the duplication of the values `/asciidoctorj/latest/`.

mojavelinux:
It certainly could be done. But this is one of those things that you set up once and just leave it, so I didn't really see the need.

mojavelinux:
Let's go for it!

PR #1017, opened 2022-05-12
Fix / + link in Writer’s Guide

toastal:
In the previous commit, I forgot to add the usually unnecessary `link:`
directive prefix, but butted up against a slash, this does not render a
link.

### DIFF
--- a/docs/asciidoc-writers-guide.adoc
+++ b/docs/asciidoc-writers-guide.adoc
@@ -2252,8 +2252,8 @@ Check out the {user-ref}[Asciidoctor User Manual] and the {docs-ref}[Asciidoctor
 === Where else is AsciiDoc supported?
 
 The easiest way to experiment with AsciiDoc is online.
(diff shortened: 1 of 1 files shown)


mojavelinux:
Thanks!

mojavelinux:
The closest match in the new docs is probably https://docs.asciidoctor.org/asciidoctor/latest/tooling/#hosted-git-repositories.

toastal:
Neat. That one doesn't feature Codeberg though.

I really appreciate this project as it really covers a lot of alternative lightweight markup deficiencies as well as made my code documentation a lot easier to write.

mojavelinux:
> I really appreciate this project as it really covers a lot of alternative lightweight markup deficiencies as well as made my code documentation a lot easier to write.

I love to hear that! That's what we're all about.

> Neat. That one doesn't feature Codeberg though.

Right. I was pointing out that it would need a similar update. All future updates should be directed there as well.

PR #1021, opened 2022-07-02
add extension: asciidoctor-comment-links

diguage:


mojavelinux:
I'd rather add this as an example in the docs. The extensions on this page are mostly about full-featured integrations, not just small customizations. Besides, this isn't really an extension so much as it is a custom Rouge adapter.

mojavelinux:
I have added this use case to the docs. See https://docs.asciidoctor.org/asciidoctor/latest/syntax-highlighting/custom/#create_formatter-rouge That's where I'd like it to stay for now.

PR #1029, opened 2023-09-26
Retrieve Tiered List of Supporters from OpenCollective API

r0ckarong:
I can't for the life of me get awestruct to work on my Windows machine. The code itself was pretty easy but I wasted about half a day trying to set up the ruby env to see if my changes would actually look right but I failed. I will try this again tonight when I get home to my trusty Linux machines. They'll comply ... or else.

I tried adapting everything as simply as possible with as little changes as I could make. Please have a look.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -261,6 +261,8 @@ task :validate do
   run_proofer
 end
 
+Dir.glob('_tasks/*.rake').each {|file| load file }
+
 # Execute Awestruct
 def run_awestruct(args, opts = {})
   cmd = "#{$use_bundle_exec ? 'bundle exec ' : ''}awestruct #{args}"
--- a/_config/supporters.yml
+++ b/_config/supporters.yml
@@ -1,263 +1,284 @@
-- :category: Change Makers # $1000/month
+---
+- :category: Change Maker
+  :members: []
+- :category: Innovation Sponsor
   :members:
-    - :name: '{ okta }'
-      :github: oktadeveloper
-      :avatar_url: https://www.okta.com/sites/all/themes/Okta/images/blog/Logos/Okta_Logo_BrightBlue_Medium.png
-      :avatar_link: https://developer.okta.com/signup?utm_source=asciidoctor&utm_medium=logo&utm_campaign=sponsor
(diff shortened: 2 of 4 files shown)


r0ckarong:
Oh great, the CI has some of the same problems installing the toolchain ... narf

r0ckarong:
Now I've gotten `some` version of the toolchain to work and now I'm banging my head against the SLIM format for awestruct.

r0ckarong:
I've managed to get the preview working but I'm still getting way too many errors:

Screenshot from 2023-09-26 22-09-34

I underestimated how much more complicated the processing in the template was than from my tests.
You're probably way faster adjusting this but feel free to reject the PR.

Preferrably with a pointer to a good example on how to do the template right :roll_eyes:

mojavelinux:
> Preferrably with a pointer to a good example on how to do the template right

It appears that you had too much indentation on certain lines. Slim is very picky about indentation and will fail if it is wrong. It was the line with the `a` tag that had too much indentation.

mojavelinux:
I updated the script to work with the existing page template, aside from a few tweaks to clean it up.

mojavelinux:
> I underestimated how much more complicated the processing in the template was than from my tests.

I think the main issue is your Ruby version. 2.4.0 is substantially out of date. I was successful using Ruby 2.7, which is documented in this file: https://github.com/asciidoctor/asciidoctor.org/blob/main/netlify.toml

r0ckarong:
> > I underestimated how much more complicated the processing in the template was than from my tests.
> 
> I think the main issue is your Ruby version. 2.4.0 is substantially out of date. I was successful using Ruby 2.7, which is documented in this file: https://github.com/asciidoctor/asciidoctor.org/blob/main/netlify.toml

I used 2.4 because that's what it says in the README. I usually have 3.1.3 or higher.

mojavelinux:
Ah, that was just out of date (no surprise since we hardly ever touch this repository these days). I updated it to match what CI is using. When it doubt, use what CI is using because that's what we know works (assuming CI isn't failing).

mojavelinux:
Thanks for the head start to get this page switched over to data from OpenCollective! I've been meaning to do this for years.

PR #1032, opened 2023-12-06
Update extensions.adoc

sanyasandor:
extended with HTML Image Map block processor

### DIFF
--- a/docs/extensions.adoc
+++ b/docs/extensions.adoc
@@ -180,4 +180,8 @@ Each cell is formatted as an AsciiDoc table cell content, so the event can inclu
 |https://github.com/Alwinator/asciidoctor-lists[Asciidoctor Lists]
 |Asciidoctor
 |An asciidoctor extension that adds a list of figures, a list of tables, or a list of anything you want!
+
+|https://launchpad.net/imagemap-block-processor[HTML Image Maps]
+|Asciidoctor
+|An asciidoctor extension that generates a html imagemap. The pre-defined areas on the image can point to chapters in the document. Feedbacks, proposals are highly appreciated.
 |====


mojavelinux:
Can you provide a syntax example in the README of the project that shows how this extension is used in an AsciiDoc document?

sanyasandor:
README is now updated. Also attached a generated html next to the example.adoc

mojavelinux:
Thanks. Though when I said README, I was actually talking about the description on this page: https://launchpad.net/imagemap-block-processor I missed that there was a README in the repository.

> Also attached a generated html next to the example.adoc

That seems unnecessary. Generally you shouldn't commit generated files.

When I looked at the README, I noticed this:

```
==  ImageMap example
:doctype: html
```

That's not valid. The doctitle should only have single `=`, and html is not a valid doctype. It should be:

```
=  ImageMap example
```

As for merging this, I would prefer to add it to this list here: https://github.com/asciidoctor/asciidoctor-extensions-lab#other-extensions There are things about this extension that I consider to be unorthodox and I don't want to promote it (at least not yet) as a community extension. The lab is really where this reference belongs for now.

sanyasandor:
Thanks for the hint. I updated that introduction page on Launchpad as well
and removed the :doctype: from all examples, also the generated file is
gone. My goal was to share this, was not sure about the proper place. This
one seemed practical with a simple link to the repository.

How shall I proceed with the extensions-lab ? Shall I make a pull request
there initially, and whenever an update is coming ?

Dan Allen ***@***.***> ezt írta (időpont: 2023. dec. 7., Cs,
20:56):

mojavelinux:
I can take care of adding it there. No need to worry. You've provided all the information I need. Thanks!

sanyasandor:
Great ! thanks ! :-)

Dan Allen ***@***.***> ezt írta (időpont: 2023. dec. 7., Cs,
21:42):

PR #1040, opened 2025-04-05
Update asciidoc-writers-guide.adoc

windupbird144:
I installed `asciidoc 10.2.1` from AUR and ran the example from the docs.

The docs claim that 
```
[.userinput]#asciidoc#
```
generates

```
<span class="userinput">asciidoc</span>
```

but in fact it generates

```
<span class=".userinput">asciidoc</span>
```

which leads me to believe you need to leave out the . in the square brackets.

### DIFF
--- a/docs/asciidoc-writers-guide.adoc
+++ b/docs/asciidoc-writers-guide.adoc
@@ -268,7 +268,7 @@ The role can be used to apply custom styling to the text.
 For instance:
 
 [source]
-Type the word [.userinput]#asciidoc# into the search bar.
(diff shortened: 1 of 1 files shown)


mojavelinux:
You are talking about the behavior of AsciiDoc Python. This website is about Asciidoctor and the pre-spec AsciiDoc that it implements. Thus, this change is not relevant in this context.

windupbird144:
You're right, sorry for the confusion! Arch users want the package [extra/asciidoctor](https://archlinux.org/packages/extra/any/asciidoctor/) for the asciidoctor tool written in Ruby.

PR #1041, opened 2025-05-10
Add asciidoctor-mathematical extension to the docs

hybras:


mojavelinux:
The reveal.js converter is not an extension, it is an extended converter. It's already listed here:

https://asciidoctor.org/docs/#extended-converters

I'm fine with adding Asciidoctor Mathematical to the third-party extension list.

mojavelinux:
The build is failing because there is trailing spaces on two of the lines.

